PR #57, opened 2020-06-12
remove pre ios 10 code

krugerk:
deployment target was bumped from 8.3/9.3 to 10.0 in Commit b4e66eaa

thus it seems we no longer need to check that the code is being
run on at least iOS 10


andrewpbrett:
This looks great, thank you! Could we get you to fill out http://beeminder.com/cla before I merge it in?

6Vmf2y48AaGbfTD2-vfDsr759VA49m-uePFkE:
done.

PR #58, opened 2020-06-12
sort goals consistently

krugerk:
goals are sorted according to the sort-goals-by preference not only in the list displayed on the gallery but also in the lists displayed on configureNotifications and healthKitConfig


andrewpbrett:
So I'm not positive that it makes sense to have the sort preference apply to goals in places besides the dashboard. I wouldn't expect a settings screen to order goals in a way that changes (or could change) from day to day. See, for example, the Reminders page on the website - https://www.beeminder.com/reminders/[username]. It does let you click any of the columns and sorts things in place, but that sort order is not persistent.

krugerk:
https://www.beeminder.com/[username] allows for sorting in the same way, by column in the same way as the reminders (which seems to default to sorting by slug) page does. However, it appears to sort by deadline upon login. And true, each are sorted independently.

I myself only have a dozen or so active goals and do not necessarily need any of the sorting options.

krugerk:
This appears to be the original feature request for sorting by name in the app: https://forum.beeminder.com/t/feature-request-alphabetical-or-time-listing-on-iphone-app/324

Issue #44, opened 2020-05-09
Be smarter about warning me to unlock my phone


dreeves:
User Kirsty says:

> It seems to remind me to unlock my phone to sync my health data with Beeminder every day no matter that I unlock my phone many many many times each day.  I don't know why this reminder is necessary but it either needs something other than it says or it's pointless.

dreeves:
Kirsty adds:

> the notification popped up this morning while I was using my phone so "unlock your phone to.." seemed very weird for it to be saying!

andrewpbrett:
I think we just need to remove this notification entirely.

krugerk:
I myself would have created a new ticket and referenced this one in the description.
This ticket was 'be smarter about the unlock' notification, especially that it not be shown while the user's device was not locked while the user was navigating in the app.
This was implemented.

The new task seems to be instead to remove the notification and its setting entirely.

To facilitate tracking changes, relevant for release notes, I would have made that its own task/ticket.

andrewpbrett:
I could see why a separate ticket might be better, but the behavior described by the user is that the notification arrives even though her phone has been unlocked during the day. Since that behavior is still happening, I think it's the same ticket. I shouldn't have closed the issue in the first place, that's my fault.

krugerk:
No worries. Perhaps you can shed some light into why the notification is needed (was needed).

andrewpbrett:
Since Apple Health data isn't available when the phone is unlocked, if someone, for example, had a steps goal that needed steps before midnight and they went for a walk from 23:30 - 00:15 in order to get the steps in, but didn't unlock the phone while they were walking, there would be no way for Beeminder (or any other app) to update the step total before midnight. 

However, even if the phone is unlocked, there's not a guarantee that the health goal will be synced. Since the notification is 1) misleading (it implies that opening your phone is all that it takes to sync the data) and 2) not working as intended (it arrives even if the phone has been unlocked recently) I think we need to remove it entirely until we can rethink it. I don't think it's a priority to reimplement, however.

krugerk:
Yeah those original motivations are good to know.

mauricioralcala:
Hello, this notification is driving me insane. The only notifications I have on in my phone are Beeminder (this ensures I take care of my commitments and focus better) and Phone app. But I get this multiple times per day, despite unlocking my phone many times. Could we get the option to turn this off?

krugerk:
[...]

PR #78, opened 2020-07-03
[#44] smarter about unlock reminder

krugerk:
we can at least avoid displaying the unlock reminder
while the app is active

### DIFF
--- a/BeeSwift/AppDelegate.swift
+++ b/BeeSwift/AppDelegate.swift
@@ -14,8 +14,7 @@ import Sentry
 import AlamofireNetworkActivityIndicator
 
 @UIApplicationMain
-class AppDelegate: UIResponder, UIApplicationDelegate {
-
+class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterDelegate {
     var window: UIWindow?
 
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplicationLaunchOptionsKey: Any]?) -> Bool {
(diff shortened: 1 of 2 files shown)


BeeSwift/AppDelegate.swift
```diff
@@ -143,4 +142,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             return goal.relativeLane.intValue < -1
         }).count
     }
+    
+    // MARK: - UNUserNotificationCenterDelegate
+    
+    func userNotificationCenter(_ center: UNUserNotificationCenter, willPresent notification: UNNotification, withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
```
andrewpbrett, line 148: This will clear out the pending notifications, but will it silence the incoming notification that triggered it? Based on [the docs](https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/1649518-usernotificationcenter) it looks like we might want to call the `completionHander` with `UNNotificationPresentationOptionNone`.

krugerk: adjusted to call the completionHandler and with an empty list of options in the case of "unlock phone" reminders while the app is active

andrewpbrett: Sorry for the holdup here - I didn't understand why you used `[]` instead of `UNNotificationPresentationOptionNone` until I tried it myself and Xcode told me to just use `[]` :)

krugerk:
this also fixes #96

PR #81, opened 2020-07-03
Update README.md

krugerk:
updated readme with description of main features


krugerk:
@andrewpbrett oh sorry, it was still in draft at that point.
I have since pushed the list of supported Apple Health metrics.

Issue #43, opened 2020-05-07
Graph images flicker on gallery screen


andrewpbrett:
## Reseminata 

From the gallery screen, go into a goal, and then back to the gallery screen. 
Or, alternatively, pull to refresh your goals

## Expectata

Only graphs that have changed should update

## Resultata

Every single graph briefly flickers the placeholder bee image.

krugerk:
This happens because as is, the goals are fetched each time the goal gallery is viewed. 

I am working on this.

krugerk:
The screenshots in the google play store for the android Beeminder app show an option for setting the interval at which the app will check for updated graphs.
It not clear to me how up-to-date those images are.
Nonetheless, is such a setting along a direction in which this app should be headed?

dreeves:
Negative, the plan is to remove that setting from Beedroid. [Anti-Settings Principle](http://doc.bmndr.com/choices)!

PR #89, opened 2020-07-10
segmentedControl for sorting goals on gallery

krugerk:
- allow sorting of the goals directly from the gallery
via the SegmentedControl
 - sort by preference is synced with one available in the
app settings
 - can combine filtering (by slug) with quick sort via segmentedControl


krugerk:
In keeping up with the previous discussion on sorting goals (MR #58), where it was decided the three lists of goals should not share a common sort but rather the sorting of goals on the gallery should be closer to what one sees on the website, a segmented control has been introduced to allow "sorting by tapping column headers".
The sorting preference in the app settings and the selected segmented control are kept in sync and only used on the gallery.

(Unrelated but still noticeable are the flickering graphs, noted Bug #43.)

krugerk:
GUI could use a bit of tweaking with dark appearance and/or on iOS 13 (with its different appearance for segmented controls -  for example the visible color behind the rounded corners of the segmented control)

andrewpbrett:
I like this idea, I think we should get some user feedback about it. I want to get 5.7 released first and then we can put this into a 5.8beta and upload to TestFlight.

krugerk:
If the sort pref only is for the gallery and we use this segmented control instead, then there need not be a separate screen just for picking the sort criteria.

PR #90, opened 2020-07-10
Gemfile and check for pristine working dir

krugerk:
Introduced a Gemfile to declare our dependency on cocoapods. Furthermore, this allows one to run pod over bundler:
```
$ bundle install
$ bundle exec pod install
```

Considering Pods are being included in the repository, we expect there to be no changes in the working directory after running `pod install` - these changes should have been added to the index and saved as a commit, with the other changes that introduced changes to the Pods folder.
A github workflow (action) has been added to check that this is the case, currently for pushes. Thus each push shall include feedback that all relevant changes have been included - by checking for a pristine working directory.


.github/workflows/clean-working-dir.yml
```diff
@@ -0,0 +1,28 @@
...
+        bundle install
+        bundle exec pod install
+    - name: lint working directory
+      run: |
+        if output=$(git status --porcelain) && [ -z "$output" ]; then
+          echo "::debug::working dir is clean"
+        else
+          echo "::error ::dirty working directory - uncommitted changes"
+          exit 1
+        fi
```
andrewpbrett, line 28: @krugerk Is the expected behavior here that this gets run before every push? I checked out this branch locally, and made a dummy commit, and pushed it to origin. It didn't seem like it ran `pod install` but it may have done so silently? The commit was just whitespace so I wouldn't have expected there to be any changes to the Podfile or lock file. 

I do like the idea of having this check, just want to make sure I understand what it's doing.

krugerk: It runs on GitHub once GitHub receives the push.

A similar pre-push hook could be used to run the same locally before sending the commits to GH.

andrewpbrett: Ah got it - and then what happens if the working directory is dirty? The push is rejected? It seems like not-good design that someone sends a push and then would have no idea that it's rejected.

krugerk:
There are two pieces. One is the Gemfile declaring that we use cocoapod. Having the Gemfile allows GitHub to check dependencies for known vulnerabilities and it allows developers to run bundle install to install gem pod and then bundle exec pod whatever (install) as one would with the pod executable. We could even set the specific version of cocoapod needed in the Gemfile as is done for Pods in the Podfile. 

The other change was that of the GitHub action. There can be many actions. One might build the app and / execute the test targets with such an action. The results of the action are then posted to the merge/pull request or push, depending on how the action is configured.
This particular GitHub action just checks that all changed files have been added. I have another one elsewhere that runs lint across the code. These could be some basic checks as part of CI but arguably could be run locally as well.
Having it server side, we can assure our master is kept clean.

Shall I split the MR into two? One for the 'working dir is clean' action along with an additional pre-push git hook and the other introducing the Gemfile?

.github/workflows/clean-working-dir.yml
krugerk: There are two pieces. One is the Gemfile declaring that we use cocoapod. Having the Gemfile allows GitHub to check dependencies for known vulnerabilities and it allows developers to run bundle install to install gem pod and then bundle exec pod whatever (install) as one would with the pod executable. We could even set the specific version of cocoapod needed in the Gemfile as is done for Pods in the Podfile.

The other change was that of the GitHub action. There can be many actions. One might build the app and / execute the test targets with such an action. The results of the action are then posted to the merge/pull request or push, depending on how the action is configured.
This particular GitHub action just checks that all changed files have been added. I have another one elsewhere that runs lint across the code. These could be some basic checks as part of CI but arguably could be run locally as well.
Having it server side, we can assure our master is kept clean.

Shall I split the MR into two? One for the 'working dir is clean' action along with an additional pre-push git hook and the other introducing the Gemfile?

Issue #92, opened 2020-07-14
Today Widget seems to prefer compact (rarely if ever shows expanded)


krugerk:
## steps to reproduce

1. have more than one goal
2. install the today widget
3. note that the beeminder today widget is shown in compact mode (just one goal) and does not offer the option to expand (for up to three goals)

## Expected

given an account with more than one goal, the today widget should offer an expanded view allowing for displaying up to three goals at a time

## Actual

today widget displayed only one goal with no option to expand it to display more

## notes

Looking over the code, it seems a quick fix might be as easy as moving setting `widgetLargestAvailableDisplayMode` to after having obtained the `goalDictionaries` (via _updateDataSource_)

## hardware / software

recent master branch (Commit d7c687222)
Simulator, iOS 13.5 / iPhone 11 Pro

PR #93, opened 2020-07-14
[#92] fix today widget's bias toward compact view

krugerk:
allowing for expanded was calculated and set before obtaining the data;
now grabbing the data/dictionaries first; if they contain more than one
entry, the expanded display mode is marked as available

### DIFF
--- a/BeeSwiftToday/TodayViewController.swift
+++ b/BeeSwiftToday/TodayViewController.swift
@@ -22,13 +22,13 @@ class TodayViewController: UIViewController {
     override func viewDidLoad() {
         NotificationCenter.default.addObserver(self, selector: #selector(self.updateDataSource), name: UserDefaults.didChangeNotification, object: nil)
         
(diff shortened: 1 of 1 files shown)


BeeSwiftToday/TodayViewController.swift
```diff
@@ -22,14 +22,14 @@ class TodayViewController: UIViewController {
     override func viewDidLoad() {
         NotificationCenter.default.addObserver(self, selector: #selector(self.updateDataSource), name: UserDefaults.didChangeNotification, object: nil)
         
+        self.updateDataSource()
+        
         self.preferredContentSize = CGSize.init(width: 0, height: self.rowHeight)
         
         if self.goalDictionaries.count > 1 {
```
krugerk: agreed it makes sense to calculate this after `updateDataSource`, just this commit is not 'defensive' enough

Issue #94, opened 2020-07-16
improve dark mode support


krugerk:
Apparently the app supports dark mode in that it has not been disabled. However, a few of the screens and image assets do not work well when in dark mode. 

## Come again, the steps to reproduce the bug from scratch:

1. grab a recent HEAD, Commit fcc79a30b8655ceb9ce66dde21d7c7cd3eef08b5 (or [Version 5.6.4](https://github.com/beeminder/BeeSwift/releases/tag/5.6.4))
2. start a simulator, configure it for dark appearance
3. launch the app
4. notice the sore spots:
   - so few of the images are visible on the settings screen
   - the launch screen contains the bee surrounded by a white box surrounded by black
   - similar for the sign in / sign up screen
 

## and what did you expect

I would expect Dark mode to be supported (work and look good) or disabled until such time.

## and what actually happened instead

As noted above, the launch screen was rather jarring. The missing icons in the settings were not so bad (one could read the label anyway) but still felt unpolished.

#### before - dark mode on goalvc with graph placeholder
![dark mode - goalvc with graphplaceholder](https://user-images.githubusercontent.com/4656811/87732760-253a7580-c7ce-11ea-915a-77e4b9ad2114.png)

#### before - dark mode on launch screen
![dark mode - launch screen](https://user-images.githubusercontent.com/4656811/87732765-266ba280-c7ce-11ea-8851-26d72e72f6ad.png)

#### before - dark mode on log in / sign up screen
![dark mode - log in or sign up screen](https://user-images.githubusercontent.com/4656811/87732767-27043900-c7ce-11ea-87e7-9d2f4554c8da.png)

#### before - dark mode on settings screen showing black icons on dark grey background
![dark mode - settings with black icons on grey background](https://user-images.githubusercontent.com/4656811/87732771-279ccf80-c7ce-11ea-96f2-aa7093adc5c6.png)

# report on how it was resolved

For images in asset catalogs, the appearances supported should be selected. This allows defining the images to be used anytime, just light, just dark. I have created 'dark mode' images for most of the images by merely inverting the black to white. The graph placeholder and thumbnail placeholder now have dark mode versions with a mostly black background.

PR #95, opened 2020-07-16
[#94] dark mode images

krugerk:
and here the proposed after screenshots:

#### goalVC
after - dark mode - goalvc

#### launch screen
after - dark mode - launch screen

#### log in or sign up
after - dark mode - log in or sign up

#### settings
after - dark mode - settings


andrewpbrett:
This looks great, nice catch especially on the settings screen.

PR #107, opened 2020-07-30
remove facebook and twitter, fix #73, fix #67

### LINKED ISSUE #67: Twitter login should work
should be able to log in with twitter

twitter button not available

### LINKED ISSUE #73: Sign in with Apple
[Try to include a use case maybe. And delete this boilerplate.]

andrewpbrett:
this removes third party login from the app, since we're now required to either implement Apple Sign In or not have any third party logins.


krugerk:
the [sample config.swift](https://github.com/beeminder/BeeSwift/blob/61dbe8f6cdfeff63d86da1b019cfcda808313f0b/BeeSwift/Config.sample.swift) can be updated, removing the placeholders for 3rd parties: twitter, google, and facebook.

PR #110, opened 2020-07-30
signin screen clean up

krugerk:


krugerk:
## Before

1. choose
1. sign in
1. sign up
1. tapped sign up although fields are blank
1. response from network request shown - cannot sign up

before-animation


## After

1. choose
1. sign in
1. sign up
1. alert showing data is missing

after-animation

PR #118, opened 2020-08-03
[47] repeat comment from previous datapoint

krugerk:
already repeating the value from the previous datapoint, now the comment too

enabled the clear button for the comment field as well

### DIFF
--- a/BeeSwift/GoalViewController.swift
+++ b/BeeSwift/GoalViewController.swift
@@ -221,6 +221,7 @@ class GoalViewController: UIViewController, UITableViewDelegate, UITableViewData
         self.commentTextField.delegate = self
         self.commentTextField.placeholder = "Comment"
         self.commentTextField.returnKeyType = .send
+        self.commentTextField.clearButtonMode = .always
         self.commentTextField.snp.makeConstraints { (make) -> Void in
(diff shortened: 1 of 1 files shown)


BeeSwift/GoalViewController.swift
```diff
@@ -475,6 +477,12 @@ class GoalViewController: UIViewController, UITableViewDelegate, UITableViewData
         }
     }
     
+    func setInitialCommentTextField() {
+        guard let lastDatapoint = self.goal?.recent_data?.last as? JSON, let comment = lastDatapoint["comment"].string else { return }
```
krugerk, line 481: easy enough
the same is done for the value

but what if one were to delete the comment each time? sure the clearButton is being shown but could mean an extra tap

andrewpbrett:
I believe @dreeves has a philosophical objection to this; namely, that we should add just a slight amount of friction to help ensure legitimate data and also not encourage duplicate comments. I could be wrong, and it could no longer be the case. The website, however, does not pre-fill the comment field. I'm going to close and let @dreeves open if he wants.

andrewpbrett:
er, sorry, I just re-read the linked issue. I was basing the comment above on the assumption that this was pre-filling the comment field. That's still a no-go, but setting placeholder text would be fine.

PR #122, opened 2020-08-03
fix #85 health app integration vc cleanup

### LINKED ISSUE #85: health app integration VC cleanup
Here are some notes on some cleanup action ideas for the screen. The health app integration screen displays what appears to be a list, starting with a switch for sync health data reminders, followed by the goals and their associated data source. The goals belong to any of three categories: auto-sourced, manually sourced, and health integration sourced. "Add source..." appears next to the goals currently manually sourced. Ones that are auto-sourced display the source greyed out. And those with a health integration display the health metric.

### Sorting of Goals

**Actual:**
Goals are sorted in reverse alphabetical order by slug. 
**Expected:**
There is no option to adjust how they are sorted on this screen. I would think sorting them A-Z would be preferred over the reverse, Z-A.

**Actual:**
Goals are sorted in reverse alphabetical order by slug. 
**Expected:**
Allow sorting the goals on this page, perhaps on-the-fly, or by using the order specified in the app preference for sorting goals (set in settings; used by the gallery)

### Use Sections
**Actual:**
Goals of the three categories are intermixed.
**Expected:**
Display section headers and split the goals into categories by type of source. The ones that cannot be edited in the app can appear toward the button, those where currently "Add source..." is displayed can either be in the first or second section. Using sections it can be made more clear which ones are sourced via apple health metric and can be changed in the app.

### pull-to-refresh
**Actual:**
There is no way to refresh the screen (no refresh button, no pull-to-refresh). refreshed by returning to the health app integration screen
**Expected:**
pull-to-refresh is supported

### cell reuse bug
**Actual:**
The greyed out source applies to wrong goals (example apple health sourced ones) upon scrolling the list
**Expected:**
The coloring is correct even after scrolling.

### greyed out
**Actual:**
The goal is not greyed out but the source is for those auto-sourced elsewhere.
**Expected:**
Both the goal and the source should be greyed out to indicate more clearly the item is not editable.

### which goal
**Actual:**
After tapping "Add source..." one is presented with "Choose metric" and a list of metrics, but no longer sees for which goal (obviously the one we tapped, but which was that again?)
**Expected:**
The screen for picking the metric could indicate for which goal. Another option might be to use a UIPickerView.

### sorting of metrics
(This applies to the list of metrics and not the health app integration screen itself.)
**Actual:**
Some of the items appear grouped together but it is not obvious how they are sorted.
**Expected:**
At a minimum, sort the health kit metrics alphabetically.

krugerk:
resolves #85

### DIFF
--- a/BeeSwift/HealthKitConfigViewController.swift
+++ b/BeeSwift/HealthKitConfigViewController.swift
@@ -18,7 +18,7 @@ class HealthKitConfigViewController: UIViewController {
     let cellReuseIdentifier = "healthKitConfigTableViewCell"
     var syncRemindersSwitch = UISwitch()
     let margin = 12
-
+    
     override func viewDidLoad() {
         super.viewDidLoad()
         if #available(iOS 13.0, *) {
@@ -71,15 +71,13 @@ class HealthKitConfigViewController: UIViewController {
     }
     
     func sortGoals() {
-        self.goals.sort { (goal1, goal2) -> Bool in
-            return goal1.slug > goal2.slug
-        }
+        self.goals.sort { $0.slug < $1.slug }
     }
     
     override func viewDidAppear(_ animated: Bool) {
         self.fetchGoals()
     }
-
+    
     override func didReceiveMemoryWarning() {
         super.didReceiveMemoryWarning()
         // Dispose of any resources that can be recreated.
@@ -112,12 +110,12 @@ class HealthKitConfigViewController: UIViewController {
         }
     }
     
-    @objc func handleMetricRemovedNotification(notification : Notification) {
+    @objc func handleMetricRemovedNotification(notification: Notification) {
         self.fetchGoals()
     }
 }
 
-extension HealthKitConfigViewController : UITableViewDelegate, UITableViewDataSource {
+extension HealthKitConfigViewController: UITableViewDelegate, UITableViewDataSource {
     func numberOfSections(in tableView: UITableView) -> Int {
         return 1
     }
@@ -128,7 +126,7 @@ extension HealthKitConfigViewController : UITableViewDelegate, UITableViewDataSo
     
     func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
         let cell = tableView.dequeueReusableCell(withIdentifier: self.cellReuseIdentifier) as! HealthKitConfigTableViewCell!
-
+        
         let goal = self.goals[(indexPath as NSIndexPath).row]
         cell!.goal = goal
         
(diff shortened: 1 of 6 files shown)


BeeSwift/HealthKitConfigViewController.swift
```diff
@@ -92,14 +115,16 @@ class HealthKitConfigViewController: UIViewController {
         }
     }
     
-    func fetchGoals() {
+    @objc func fetchGoals() {
         MBProgressHUD.showAdded(to: self.view, animated: true)
         CurrentUserManager.sharedManager.fetchGoals(success: { (goals) in
             MBProgressHUD.hideAllHUDs(for: self.view, animated: true)
             self.goals = goals
+            self.tableView.refreshControl?.endRefreshing()
```
krugerk, line 123: a tweak would be addressing having both MBProgressHUD and refreshControl visible

BeeSwift/HealthKitConfigViewController.swift
```diff
@@ -112,31 +137,59 @@ class HealthKitConfigViewController: UIViewController {
...
+        if section == 0 {
+            return "manually beeminded goals"
+        } else if section == 1 {
+            return "modifiable, Apple Health goals"
+        } else {
+            return "unmodifiable Auto-data goals"
+        }
+    }
+    
+    func goalAt(_ indexPath: IndexPath) -> JSONGoal {
```
krugerk, line 181: not dealing with OutOfRange directly here

`goalAt` can be private

BeeSwift/RemoveHKMetricViewController.swift
```diff
@@ -53,14 +54,20 @@ class RemoveHKMetricViewController: UIViewController {
     @objc func removeButtonPressed() {
         guard self.goal != nil else { return }
         self.goal?.autodata = ""
-        var params : [String : [String : String]] = [:]
-        params = ["ii_params" : ["name" : "", "metric" : ""]]
+        var params: [String: [String: String]] = [:]
```
krugerk, line 57: since we touched these lines (formatting), just as well merge the two into `let params`

BeeSwift/RemoveHKMetricViewController.swift
```diff
@@ -24,7 +25,7 @@ class RemoveHKMetricViewController: UIViewController {
         
         let currentMetricLabel = BSLabel()
         self.view.addSubview(currentMetricLabel)
-        currentMetricLabel.text = "This goal gets its data from Apple Health (\(self.goal!.humanizedAutodata()!)). You can disconnect the goal with the button below."
+        currentMetricLabel.text = "This goal (\(goal.slug)) obtains its data from Apple Health (\(self.goal.humanizedAutodata!)). You can disconnect the goal with the button below."
```
krugerk, line 28: yay, we see again from which goal the health metric would be disconnected

krugerk: ideally we would see the same when adding as well

PR #124, opened 2020-08-06
add fastlane

andrewpbrett:


krugerk:
Would you like to write [fastlane lanes and such in swift](https://docs.fastlane.tools/getting-started/ios/fastlane-swift/)? That would explain the fastlane/swift folder but from what I can tell the project is configured in the standard ruby version (fastlane/Fastile)

krugerk:
I removed them in the 'clean up' PR - #140

andrewpbrett:
Ah thank you, I did try doing them in Swift first but it wasn't working and I then I noticed that it was in beta. Fine to remove that folder.

PR #126, opened 2020-08-07
repo cleanup remove googlesignin bundle

krugerk:


krugerk:
> double check, looks good

Oh, file is still mentioned in the git ignore file.

PR #128, opened 2020-08-10
use SafariViewController for when viewing goals externally

krugerk:
similar to #119 
use the safariVC to view the goal
this way the user "stays in the app" (or at least can more easily return when done viewing/editing the goal via the website)


krugerk:
> This looks good - I did notice that the Safari VC scrolls down to the data entry box (since it has focus) which might be a little jarring. I don't know if this happened with the old method too, or if it would be difficult to change. Not needed before merging, just a nice to have.

in my tests this was the behavior before as well when launching externally via UIApplication.shared.open(url, options, completed) (or ...openURL(url)

PR #132, opened 2020-08-12
use `me` instead of username when using access token

krugerk:
according to the [API documentation](http://api.beeminder.com/#client-oauth),
the access token uniquely identifies the user and thus the placeholder can be used
instead of the username

replaced username with me in urls using access token


andrewpbrett:
We actually prefer having the username since that makes it easier to look through logs for requests from a particular user; I don't think the API docs need to be updated to mention this since it's only relevant for the iOS/Android apps.

krugerk:
I understand.

PR #137, opened 2020-08-15
be consitent: use username in api call

krugerk:
a follow-up to the discussion in #132 
replacing the remaining two accessToken based calls using placeholder _me_ with the `username` instead

### DIFF
--- a/BeeSwiftToday/TodayTableViewCell.swift
+++ b/BeeSwiftToday/TodayTableViewCell.swift
@@ -163,7 +163,7 @@ class TodayTableViewCell: UITableViewCell {
     }
     
     func pollUntilGraphUpdates() {
-        if self.pollTimer != nil { return }
+        guard self.pollTimer == nil else { return }
         self.pollTimer = Timer.scheduledTimer(timeInterval: 2, target: self, selector: #selector(self.refreshGoal), userInfo: nil, repeats: true)
     }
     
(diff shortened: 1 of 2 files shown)


BeeSwiftToday/TodayTableViewCell.swift
```diff
@@ -73,12 +73,16 @@ class TodayTableViewCell: UITableViewCell {
...
+        let addTitle: String = {
+            // small screen hack
+            if (UIDevice.current.orientation == .landscapeRight || UIDevice.current.orientation == .landscapeLeft && UIScreen.main.bounds.height >= 375) ||
+                (UIScreen.main.bounds.width >= 375) {
+                return "Add data"
+            } else {
+                return "Add"
+            }
+        }()
+
```
andrewpbrett, line 85: Were these changes supposed to be included in this PR?

krugerk: > Were these changes supposed to be included in this PR? 

Obviously do not address the 'username' instead of 'me' but probably a cleanup commit along the way, as part of the 

> Uncle Bob talks about the boy scout rule, which can be summarized as:
> **Leave the code cleaner than you found it.**
> Meaning that every time that we make a change in our code base, make an improvement in the code.


It is probably its own commit here, easy to revert / remove if it need be elsewhere.

krugerk: Commit 1a01dddcf6dfc7531319b0927d642fad7237bc55

PR #138, opened 2020-08-15
elaborate on contributing guidelines

andrewpbrett:


CONTRIBUTING.md
```diff
@@ -0,0 +1,29 @@
+## Contributing
+
+#### Legal Schmegal
+
+First, please read and sign [beeminder.com/cla](https://beeminder.com/cla).
```
krugerk, line 5: This is for contributing code , right, but one could also contribute via feature requests / bug reports without. Correct?

PR #140, opened 2020-08-17
a few rewrites, avoiding force uwrap

krugerk:


krugerk:
Hi @andrewpbrett 

Understood

Could you mark the rest of the things slated for 5.7 as such?

Issue #23, opened 2019-12-13
healthkit integration info stored by goal name (and not getting cleared when a goal's deleted)


bsoule:
A user created a goal and connected to apple health.

Then they changed their mind, deleted the goal, and made a new one with the same name.

The app then continued to send healthkit data (that they didn't want) to the new goal.

(possibly sort of a problem with the beeminder api and that we use goal slug as the primary identifier... anyhow. does removing healthkit info from a goal work now? perhaps if that'd worked as expected they mightn't have deleted the original goal?)

6Vmf2y48AaGbfTD2-vfDsr759VA49m-uePFkE:
As the primary ID? I understand the slug is to be unique for a given user and this seems to be the case (cannot create a goal using the same slug as an archived goal, for example).
Of course, as you point out, once the goal is deleted, another could be created using that same name.
Does the data about a goal include something such as 'goal creation date/timestamp' or some other unique element (uuid, incremental numeric id)?
Perhaps a combination of these could be used by the app to distinguish between two goals with the same name.

krugerk:
If this is an app-side only issue, I could image the following scenario:
 
> User creates a goal and eventually uses the iOS App to link that goal with a HealthKit metric. As is ([v 5.6.4](https://github.com/beeminder/BeeSwift/releases/tag/5.6.4)), each time the app shows the gallery (and in a few other instances) all goals for a user are fetched. Upon successfully downloading the data representing the goals, the app configures any HeathKit integrations for the goals. It would be at this point that a "update goal with data from HK metric" event is scheduled with HK.
> If in the meantime the use deleted said goal replacing it with another under the same slug and did not load the app (the gallery with the list of goals), then the app would not reconfigure HK according to the new(ly updated) goals. Thus it could perhaps happen that HK triggers with new data, passing it along to a goal of matching slug, which is now a new goal.

I am not quite to the point where I am testing the HK integration thoroughly, I just noticed something similar having recently removed a goal (via the website) and created a new one (named the same, via the website). In my case, I had the goal screen open showing the old goal and the app was happy to send new data (entered at that screen) to the new goal of the same slug although it was a new goal.
It might not hurt for the API to handle this - pass around data corresponding to unique goal IDs and not just slugs.

andrewpbrett:
Another report of this one: 

[...]

PR #142, opened 2020-08-18
check for signed in before sending datapoint; fix #109

### LINKED ISSUE #109: Crash sign out sign in back
Sounds like this is an instance of #414.

5.7 (2) TestFlight

Signed in already

Signed out (to see the screens without Twitter Facebook)

Tapped already have account

Then tapped back 

And crash

Sentry: BEESWIFT-JY

andrewpbrett:


krugerk:
When signing out
Then back in
(Even after the app crashes)
Upon signing back in, the app is not redirecting to Health to enable the app.

BeeSwift/JSONGoal.swift
```diff
@@ -614,11 +614,12 @@ class JSONGoal {
     func updateBeeminderWithValue(datapointValue : Double, daystamp : String, success: (() -> ())?, errorCompletion: (() -> ())?) {
         
         if datapointValue == 0  { return }
```
krugerk, line 616: While you are here, how about changing this line to a `guard` as well?

BeeSwift/JSONGoal.swift
```diff
@@ -614,11 +614,12 @@ class JSONGoal {
...
+        guard let username = CurrentUserManager.sharedManager.username else { return }
```
krugerk, line 617: If there is no username, does that mean we are logged out? If we are logged out, shall we then just cancel pending queries and background deliveries? Would that be something to do here in the else block? Or rather call explicitly on goals like is done for setting up HK integration?

krugerk: How about we create a bug ticket for deregistering from healthkit - on sign out or when goals change such that a previous query is no longer necessary - and allow the guard here as a fix

BeeSwift/JSONGoal.swift
```diff
@@ -614,11 +614,12 @@ class JSONGoal {
     func updateBeeminderWithValue(datapointValue : Double, daystamp : String, success: (() -> ())?, errorCompletion: (() -> ())?) {
```
krugerk, line 615: As the code is written, we should probably be checking here too that this goal is still associated with a healthkit metric (and even the same one for which the query was created). This might be relevant to the issue reported where a newly created goal of the same name as one previously seen in the app was provided with HK data, apparently from the previous association (in the deleted goal of the same slug) - #23

BeeSwift/JSONGoal.swift
krugerk: created the new bug ticket as #146

andrewpbrett:
I'm going to close this PR so we can handle it properly when we do #146

PR #143, opened 2020-08-18
update fastlane to upload sentry scripts on build

andrewpbrett:


fastlane/Fastfile
```diff
@@ -21,8 +21,11 @@ platform :ios do
     run_tests(scheme: "BeeSwiftUITests")
   end
   lane :build do
-    sync_code_signing
     build_app(scheme: "BeeSwift")
+    sentry_upload_dsym(
+      org_slug: 'apb-p7',
```
krugerk, line 26: How about using env vars for these two?

https://docs.fastlane.tools/advanced/other/#environment-variables

andrewpbrett: done!

PR #145, opened 2020-08-18
add build block to semaphore config

andrewpbrett:


.semaphore/semaphore.yml
```diff
@@ -18,7 +22,28 @@ blocks:
...
+          - checkout
+          - cache restore
+          - bundle install --path vendor/bundle
+          - cache store
+          - cp /Users/semaphore/BeeSwift/BeeSwift/Config.swift.sample /Users/semaphore/BeeSwift/BeeSwift/Config.swift
+      jobs:
+        - name: Build
+          commands:
+            - bundle exec xcversion select 11.3.1
+            - bundle exec fastlane test
```
krugerk, line 42: besides the env_var section being placed elsewhere, the content of this job/task seems to be the same as the other. What does 'fastlane build' provide that 'fastlane test' does not?

andrewpbrett: that was a typo :) fixed now

PR #152, opened 2020-08-20
fix #151 overlapping views in today widget

### LINKED ISSUE #151: Views in Today Widget can be misaligned
```[tasklist]
### Desiderata
- [ ] Confirm this replicates in modern BeemiOS
```

This was noticed in the simulator and on a device running iOS 13.6 using version 5.7 (4) of the app. As seen in the screenshot, some of the text overlaps the the add-data entry area. This perhaps is a recent bug since only recently the today widget displays more than one goal (#92).

A guess is that this is a cell re-use bug because the position of the text in the second and third cells is the same as in the first - centered vertically. That is OK in the first because it is an auto goal and thus does not display the add data view elements.

overlapping views in today widget

## Cognata

* #3  

_**Verbata**: css, alignment, layout, visual bugs,_

krugerk:
fixes #151 

This merge request contains two commits dealing with two different ways of addressing the issue. Feel free to inspect both and provide feedback on a preferred route.

### DIFF
--- a/BeeSwiftToday/TodayTableViewCell.swift
+++ b/BeeSwiftToday/TodayTableViewCell.swift
@@ -33,7 +33,9 @@ class TodayTableViewCell: UITableViewCell {
     
     override func prepareForReuse() {
         super.prepareForReuse()
-        
+
+        self.subviews.forEach({$0.snp.removeConstraints()})
+
         self.valueLabel.text = nil
         self.limitLabel.text = nil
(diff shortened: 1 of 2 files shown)


andrewpbrett (changes requested):
I like the approach with two reuse identifiers.

andrewpbrett:
Is this ready to go, or still a draft?

krugerk:
Was in draft because I thought I could create some UI tests for it.

PR #154, opened 2020-08-21
Bugfix/30 graphs remain hidden after user updates payment method

### LINKED ISSUE #30: Graphs remain hidden after user updates payment method

krugerk:
fixes #30 

Rather than just fetching goals after signing in, the user resource is also fetched. The [user resource](https://api.beeminder.com/#user) contains details on 'deadbeat', with the status being updated internally once obtained.

Without complicating matters, the fetchUser is being called each time fetchGoals is called. After all, the deadbeat status can change at any time.

### DIFF
--- a/BeeSwift.xcodeproj/project.pbxproj
+++ b/BeeSwift.xcodeproj/project.pbxproj
@@ -10,6 +10,8 @@
 		1B8454411999895F722BAE80 /* Pods_BeeSwiftToday.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 8A2D58FC1D496CC59C60A22C /* Pods_BeeSwiftToday.framework */; };
 		6E72E1F82DDC41E7CD79BEE4 /* Pods_BeeSwiftTests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 20121F7A17F740656ADD55D0 /* Pods_BeeSwiftTests.framework */; };
 		9B8CA57D24B120CA009C86C2 /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 9B8CA57C24B120CA009C86C2 /* LaunchScreen.storyboard */; };
+		9B95A44C24DF8E8D007DD992 /* JSONUser.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9B95A44B24DF8E8D007DD992 /* JSONUser.swift */; };
+		9B95A44D24DF92FA007DD992 /* JSONUser.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9B95A44B24DF8E8D007DD992 /* JSONUser.swift */; };
 		A10D4E931B07948500A72D29 /* DatapointsTableView.swift in Sources */ = {isa = PBXBuildFile; fileRef = A10D4E921B07948500A72D29 /* DatapointsTableView.swift */; };
 		A10DC2DF207BFCBA00FB7B3A /* RemoveHKMetricViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = A10DC2DE207BFCBA00FB7B3A /* RemoveHKMetricViewController.swift */; };
 		A11488BE1EE9B0CE003316E1 /* UIDevice.swift in Sources */ = {isa = PBXBuildFile; fileRef = A11488BD1EE9B0CE003316E1 /* UIDevice.swift */; };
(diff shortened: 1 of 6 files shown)


krugerk:
> I'm not sure this is the best approach here, since it means every time we do a `fetchGoals` (which is pretty frequently) it's going to be two requests to the Beeminder server instead of one. For the 99.9% of users who haven't changed their deadbeat status lately, it could mean a noticeable impact on how quickly the goals appear. 
> 
> 
> 
> What about only checking if the user's already a deadbeat? 


This change set deals with just the issue mentioned. And you are right - the app downloads the full set of goals often and more often than needed. The API suggests checking user for last updated and downloading goals if needed based on that. Doing so was part of the original code change but I trimmed down to just the part with deadbeat status here.

The part about downloading goals only if user last updated is more recent than we have seen could be re-added here or in its own PR. 
How about I do that? Its own PR which targets this branch and they can eventually ride into master together.

(I had the change locally for a while and only ever noticed the initial request - the gallery empty with the message I had no goals stayed up a bit longer than usual. Considering though less being downloaded and now background fetch working, and I knew the change suppose, I was ok with it.)

krugerk:
> I'm not sure this is the best approach here, since it means every time we do a `fetchGoals` (which is pretty frequently) it's going to be two requests to the Beeminder server instead of one. For the 99.9% of users who haven't changed their deadbeat status lately, it could mean a noticeable impact on how quickly the goals appear.
> 
> What about only checking if the user's already a deadbeat?

I have opened ticket #177 describing a desired side effect of fetching user before fetching goals.

krugerk:
@andrewpbrett this Merge Request has been updated with changes for #177 to use fetching User to our advantage to not need to fetch Goals so often. Cheers.

andrewpbrett:
I might be missing something, but doesn't this still do two requests every time we call fetchGoals? One to the user endpoint, and then in the success block to the goals endpoint?

krugerk:
With the most recent commit, each fetchGoals instead calls fetchUser and only calls fetchGoals if it would contain new data.
It does this by comparing user resource's updated_at.

So no, only two calls when there are updates to the goals.

krugerk:
@andrewpbrett is the issue that the ticket described an issue with an outdated 'deadbeat' status and the appearance of making two API calls to the backend just to update deadbeat status seems like overkill?

I agree that that would be too much and apparently for something not so important (updating deadbeat status within the app).

The changes here mostly came from another branch where I users the user updated_at to determine if making a call to the goals made sense (that they had something new).
Part of those changes were updating the deadbeat status considering it would be obtained relatively often (as with everything else in the user resource).

If instead we only wish to concern ourselves with just the deadbeat status, other approaches could work: the backend could send out a push message that the status has changed (triggering an update in the app), the app could request that specific item at a much less frequent interval - once a day, once a week, ..., upon opening the settings screen, ... 

As we know, the deadbeat status is currently only set once per signin. 


If I read between the lines, my impression is you are concerned there are two calls made to the backend for something small like deadbeat status and those calls are made many times throughout typical use of the app. 

I guess I may appreciate more directed, specific ideas or suggestions. 

 > @krugerk , I see what you have done here. It might be nice to fetch goals only when there is new data but I would prefer that to be its own ticket. Instead, for this ticket, how about introducing  changes to update the deadbeat status on ever seventh app launch?

andrewpbrett:
> If I read between the lines, my impression is you are concerned there are two calls made to the backend.  

Sorry if it wasn't clear on my feedback. I was trying to being pretty clear about it in [this comment](https://github.com/beeminder/BeeSwift/pull/154#pullrequestreview-478971425), no reading in between the lines required :) 

> I guess I may appreciate more directed, specific ideas or suggestions.  

I also suggested in that comment that we limit the scope of these changes to the case where the user is a deadbeat. If user is marked as a deadbeat, it's fine to check as often as we want to see if that's changed. The changes as proposed here, again, introduce two requests to a very common operation (the case where there is new data about goals to be fetched). If people are on slower connections, this can be especially frustrating, so we can't dismiss it just because we didn't notice a delay when we tested it (probably over pretty-fast-wifi).  

I see what you're saying that if we implement the changes in #177, it will cut down on API calls overall. The app previously used this strategy and also persisted goals locally. There were a number of nightmare-class bugs related to keeping the app and the website in sync, however, so the app now uses the current approach.  

The app unfortunately isn't in a state right now where we can make a big change like this (#177) without the strong possibility of introducing bugs. I'm learning that even if the code is open source and someone can build the app, it's not possible to truly collaborate on large changes in an open source way without a more robust test suite. So building that out needs to be the first priority before accepting or starting any new (non-trivial) PR's.  

As a matter of process, I didn't notice that you'd created #177, nor that you'd added to the ToDo column in the project board and the 5.8 milestone. I'm sorry that it went unnoticed for a week and that I didn't indicate to you sooner the reasons for not wanting it to be worked on right now (see again the lesson above about not being able to effectively collaborate without tests). I'm going to try to think of a way that we can use milestones and project boards better. For now, could you hold off on adding things to milestones/projects? (Except for what happens through the automation, those are fine).

Lastly, your previous comment has a blockquote. I think you were trying to give an example of the feedback you were hoping to receive from me. I don't think it's appropriate to blockquote something that I didn't say (or to dictate verbatim what feedback you think you should have received). Would you mind removing that part or striking it out, or somehow indicating more clearly what it is?

PR #160, opened 2020-08-22
Provide Any images and not just light + dark

### LINKED ISSUE #158: Missing Settings navbaritem (icon) on GoalVC
the setting icon on the top right corner is missing from sight! I can still navigate to setting by touching that area but it’s weird to miss the icon.

iphone 6 running iOS 12.4.8
app version 5.7, app store

ac66a38f5cb9a029516ffebfe056594ba0585d1a

(from the [beeminder forum](https://forum.beeminder.com/t/nailing-down-the-changelog-for-beemios-version-5-7/7155/8))

krugerk:
fixes #158 settings icon missing on ios12

### DIFF
--- a/BeeSwift/Images.xcassets/Clock.imageset/Contents.json
+++ b/BeeSwift/Images.xcassets/Clock.imageset/Contents.json
@@ -8,16 +8,6 @@
       "appearances" : [
         {
           "appearance" : "luminosity",
-          "value" : "light"
-        }
-      ],
-      "idiom" : "universal",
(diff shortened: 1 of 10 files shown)


krugerk:
I had this set to draft as I experimented with UITesting to check which image was being used. From what I could tell, it was possible to check which view was loaded but not which image.
Another approach might be to compare screenshots.

krugerk:
Issue was introduced as part of #95, introduction of some images for dark mode

PR #163, opened 2020-08-23
support in-app sign-up

### LINKED ISSUE #162: App crashes when Signing Up in the app
1. fresh install, 5.7
2. at sign in / sign up, tap 'Create a Beeminder Account'
3. enter desired username, desired password, and email address
4. tap sign up
5. found nil while unwrapping

Since offered, one can use the app to sign up for an account via email address.

The account was created but the app did not proceed with it.

Instead

> Fatal error: Unexpectedly found nil while unwrapping an Optional value: 

# Notes

OAuthSignInManager.signUpWith(email
-> CurrentUserManager handleSuccessfulSignin

where a responseJSON["access_token"] is requested but does not exist in the response.

Instead the response contains an authentication_token and username
```
  "authentication_token" : "some auth token",
  "username" : "newly created account's username"
```

The difference between auth token and access token as mentioned https://api.beeminder.com/#auth.

krugerk:
replace accesstoken with token (apitoken) and support both authToken and accessToken

fixes #162

### DIFF
--- a/BeeSwift/CurrentUserManager.swift
+++ b/BeeSwift/CurrentUserManager.swift
@@ -22,6 +22,14 @@ class CurrentUserManager : NSObject {
     static let goalsFetchedNotificationName = "com.beeminder.goalsFetchedNotification"
     static let healthKitMetricRemovedNotificationName = "com.beeminder.healthKitMetricRemovedNotification"
     
+    fileprivate struct UserDefaultsKeys {
+        static let apiTokenType = "api-token-type"
+        static let apiTokenValue = "api-token-value"
+        
+        static let accessToken = "access_token"
+    }
+    
+    fileprivate let apiTokenKey = "api_token_key"
     fileprivate let accessTokenKey = "access_token"
     fileprivate let usernameKey = "username"
     fileprivate let deadbeatKey = "deadbeat"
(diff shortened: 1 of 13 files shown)


BeeSwift/CurrentUserManager.swift
```diff
@@ -105,11 +118,25 @@ class CurrentUserManager : NSObject {
...
+        guard let authToken = responseJSON["authentication_token"].string, let username = responseJSON["username"].string else {
+            return
+        }
+        
+        CurrentUserManager.sharedManager.setApiToken(ApiToken(type: .AuthenticationToken, token: authToken))
+
+        UserDefaults.standard.set(username, forKey: usernameKey)
+        UserDefaults.standard.synchronize()
+        
+        NotificationCenter.default.post(name: Notification.Name(rawValue: CurrentUserManager.signedInNotificationName), object: self)
```
krugerk, line 131: After successfully signing UP, what about next signing IN with that data?

The sign in response contains data about default notifications.

krugerk: or fetching user, rather

andrewpbrett:
This looks good, but could we use this as a first opportunity to write some failing tests and make them green? Just around the sign up/sign in flow for now.

krugerk:
> This looks good, but could we use this as a first opportunity to write some failing tests and make them green? Just around the sign up/sign in flow for now. 

Been working on it - putting things in place to mock the server's responses.

andrewpbrett (changes requested):
This looks good but we'd need to have tests for the various operations that use the API token before it can be merged.

PR #171, opened 2020-08-27
Conciser HealthKit autodata instructions

### LINKED ISSUE #170: More concise HealthKit autodata instructions
## Current

Associate a HealthKit metric with your goal.
While configured as the source for this goal, Apple Health will periodically provide Beeminder for iOS with the corresponding data. Beeminder for iOS will add that data as datapoints to your goal automatically.
Tap a metric from the list below, then tap save to update your goal.

## Proposed

HealthKit metric as autodata!
With this metric as the autodata source for your goal, Apple Health will keep Beeminder up to date with your latest data automatically.
Tap a metric from the list below, then tap save to update your goal.

## New version after discussion in PR

Apple Health metric as autodata!
With this metric as the autodata source for your goal, Apple Health will keep Beeminder up to date with your latest data automatically.
(Technically it updates the Beeminder iOS app which updates Beeminder so you'll have to open the app occasionally.)
Pick a metric from the list below, then hit save to update your goal.

---

While we're at at let's conform to the Beeminder-wide spelling of "autodata" (ie, no space).

dreeves:
Closes #170

### DIFF
--- a/BeeSwift/ChooseHKMetricViewController.swift
+++ b/BeeSwift/ChooseHKMetricViewController.swift
@@ -36,7 +36,7 @@ class ChooseHKMetricViewController: UIViewController {
             attrString.append(NSMutableAttributedString(string: "\(self.goal.slug)\n",
                 attributes: [NSAttributedStringKey.font: UIFont.beeminder.defaultBoldFont]))
             
-            attrString.append(NSMutableAttributedString(string: "Associate a HealthKit metric with your goal.\nWhile configured as the source for this goal, Apple Health will periodically provide Beeminder for iOS with the corresponding data. Beeminder for iOS will add that data as datapoints to your goal automatically.\nTap a metric from the list below, then tap save to update your goal.",
+            attrString.append(NSMutableAttributedString(string: "HealthKit metric as autodata!\nWith this metric as the autodata source for your goal, Apple Health will keep Beeminder up to date with your latest data automatically.\nTap a metric from the list below, then tap save to update your goal.",
                                                         attributes: [NSAttributedStringKey.font: UIFont.beeminder.defaultFontLight.withSize(Constants.defaultFontSize)]))
             return attrString
         }()
--- a/BeeSwift/ChooseHKMetricViewController.swift
+++ b/BeeSwift/ChooseHKMetricViewController.swift
@@ -36,7 +36,7 @@ class ChooseHKMetricViewController: UIViewController {
             attrString.append(NSMutableAttributedString(string: "\(self.goal.slug)\n",
                 attributes: [NSAttributedStringKey.font: UIFont.beeminder.defaultBoldFont]))
             
-            attrString.append(NSMutableAttributedString(string: "HealthKit metric as autodata!\nWith this metric as the autodata source for your goal, Apple Health will keep Beeminder up to date with your latest data automatically.\nPick a metric from the list below, then hit save to update your goal.",
+            attrString.append(NSMutableAttributedString(string: "Apple Health metric as autodata!\nWith this metric as the autodata source for your goal, Apple Health will keep Beeminder up to date with your latest data automatically.\n(Technically it updates the Beeminder iOS app which updates Beeminder so you'll have to open the app occasionally.)\nPick a metric from the list below, then hit save to update your goal.",
                                                         attributes: [NSAttributedStringKey.font: UIFont.beeminder.defaultFontLight.withSize(Constants.defaultFontSize)]))
             return attrString
         }()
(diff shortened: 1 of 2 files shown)


krugerk:
I was not sure it would be apparent that the Apple Health sync does not talk with beeminder(.com) directly but does so only through the iOS App.

This was why the iOS app itself was mentioned in the previous text. There is currently no lengthy description about the "Sync Health data reminders", just the title. This "sync health reminder" ties into this same theme.

iOS device with Apple Health, data in Apple Health, this particular Beeminder for iOS installation on this particular iOS device, permission from Health for Beeminder to read the data (ticked off per category/metric), ... and apparently running Beeminder every now and again (for the settings to stay up-to-date, but Health should provide Beeminder the app with the data at intervals, and Beeminder the app uploads those).

BeeSwift/ChooseHKMetricViewController.swift
```diff
@@ -36,7 +36,7 @@ class ChooseHKMetricViewController: UIViewController {
             attrString.append(NSMutableAttributedString(string: "\(self.goal.slug)\n",
                 attributes: [NSAttributedStringKey.font: UIFont.beeminder.defaultBoldFont]))
             
-            attrString.append(NSMutableAttributedString(string: "Associate a HealthKit metric with your goal.\nWhile configured as the source for this goal, Apple Health will periodically provide Beeminder for iOS with the corresponding data. Beeminder for iOS will add that data as datapoints to your goal automatically.\nTap a metric from the list below, then tap save to update your goal.",
+            attrString.append(NSMutableAttributedString(string: "HealthKit metric as autodata!\nWith this metric as the autodata source for your goal, Apple Health will keep Beeminder up to date with your latest data automatically.\nTap a metric from the list below, then tap save to update your goal.",
```
krugerk: While we are here, we could consider HealthKit -> Health (the app shows up as Health and not HealthKit; HealthKit is the name of the toolkit the developers use to link to Health/its data).
What do you think?

krugerk:
```suggestion
            attrString.append(NSMutableAttributedString(string: "Apple Health as autodata!\nWith this metric as the autodata source for your goal, Apple Health will keep Beeminder up to date with your latest data automatically.\nTap a metric from the list below, then tap save to update your goal.",
```

dreeves: Agreed! I didn't realize that.

dreeves:
As for clarifying that Apple Health sends data to the app which then syncs with the server... Unless I'm confused, users don't need to know or care about that distinction. Apple Health automatically updates "Beeminder", where "Beeminder" means both the app and the website.

krugerk:
> As for clarifying that Apple Health sends data to the app which then syncs with the server... Unless I'm confused, users don't need to know or care about that distinction. Apple Health automatically updates "Beeminder", where "Beeminder" means both the app and the website.

With the other autodata sources, the user need not do anything besides enabling the interaction (beeminder(.com) -> strava, clozemaster, duolingo, so on). In those cases, Beeminder the service takes care of syncing the data. In the case of Apple Health, the user's iOS device is the necessary middleman. One needs to configure Apple Health as the source for a goal on some iOS device from within the Beeminder app. And keep the app installed. And continue to allow Beeminder access to the relevant Health data. 
Perhaps you are right and it is not at all necessary to mention any of this. However since I see this as a main feature of the iOS App (tapping into Apple Health data), it made sense to me to mention this specifically.

dreeves:
That makes sense! Let me add something...

dreeves:
How's it sounding now? (I updated gissue #170 with the latest version if it's easier to see there)

krugerk:
I am not so deep into the HealthKit stuff 
but what I think I understand from the code and documentation is that after configuring a goal with a HealthKit metric, one need not run the app to receive the updates nor to upload them.

The app uses HKQueryObserver.
These are 
"Long-running queries. These queries continue to run in the background and update your app whenever changes are made to the HealthKit store."

We can ask @andrewpbrett about the need for the sync health reminders (where one is told to open the app occasionally).

krugerk:
Re "autodata"

I used the line about not being able to change autodata sources via the website from the website as inspiration:

"At the moment we don't have a way for you to swap data source here yourself for auto-data goals. But won't it be cool when we do have a handy way to do that?"

dreeves:
Oh ho! Thanks for finding that! I've now made a corresponding PR to make the website get on board with the "autodata" spelling as well: https://github.com/beeminder/beeminder/pull/1641

That'll be great to clarify if the parenthetical warning I added -- "(Technically it updates the Beeminder iOS app which updates Beeminder so you'll have to open the app occasionally.)" -- is necessary. (Related bug report from Faire today in support about the reminder to unlock your phone.)

krugerk:
(Technically it updates the Beeminder iOS app which updates Beeminder so you'll have to open the app occasionally.)


I think the app would upload the data without the user explicitly opening the app. From what I can tell the app does not wait for that - does not build up data from health waiting to be opened by the user in order to upload the data.

PR #174, opened 2020-08-31
reset password link

krugerk:
introduces a link to the website where the user can request a password reset

The color scheme used throughout the app is used here. Thus the blue-on-black mentioned is #55 appears on the sign in view in dark mode ('Reset your password' in blue is presented on a black background). 

Fixes #173 

Also fixes the issue of the missing divider on 'sign in' that was shown on 'sign up'.

PR #175, opened 2020-08-31
replaces #174

andrewpbrett:
replaces #174

### DIFF
--- a/BeeSwift/SignInViewController.swift
+++ b/BeeSwift/SignInViewController.swift
@@ -8,8 +8,9 @@
 
 import Foundation
 import MBProgressHUD
+import SafariServices
 
-class SignInViewController : UIViewController, UITextFieldDelegate {
+class SignInViewController : UIViewController, UITextFieldDelegate, SFSafariViewControllerDelegate {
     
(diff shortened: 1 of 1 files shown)


krugerk:
@andrewpbrett I had the option _Allow edits and access to secrets by maintainers_ checked so you should have been able to modify the preview pull request without needing to create a new one. (Maybe that is easier to do with a source branch inside this project beeswift rather than from a fork.)

BeeSwift/SignInViewController.swift
```diff
@@ -25,6 +26,26 @@ class SignInViewController : UIViewController, UITextFieldDelegate {
     var backToSignUpButton = BSButton()
     var signInButton = BSButton()
     var divider = UIView()
+    let resetPasswordButton: BSButton = {
+        let button = BSButton()
+        button.setTitle("Forgot password?", for: .normal)
+        button.titleLabel?.font = UIFont.beeminder.defaultFontHeavy
+        if #available(iOS 13.0, *) {
```
krugerk, line 33: I probably would have at least grouped these `if ios13 else` together, setting both button title and background color in the same block.

PR #186, opened 2020-09-04
explicit caching of thumbnails and graphs

### LINKED ISSUE #180: Graph in goalVC should be visible right away
The scenario is gallery the tapping a goal, which loads that goal's view, then waiting while the graph loads, returning back to the gallery, and subsequently tapping and returning back to that same goal.

As is, each time the goal view is shown, first the placeholder graph is shown. I would expect this to be the case maybe the first time but not every time thereafter.

It may be the case that the image exists in the local cache but it is being set only after the placeholder is shown.

I would say the app could be more deliberate with local caching of downloaded graphs. As long as memory / iOS allows, we could be caching each of them for use everywhere until they have been updated.

Similarly, with these changes, I would expect downloaded, up-to-date thumbnail graphs in the gallery to not need to be downloaded again merely by scrolling or switching back to gallery from another view (such as a goal or in app settings).

krugerk:
Explicitly fill the cache with thumbnail and graph images and look to the cache for those images

fixes #180

### DIFF
--- a/BeeSwift/GalleryViewController.swift
+++ b/BeeSwift/GalleryViewController.swift
@@ -12,6 +12,8 @@ import MBProgressHUD
 import SwiftyJSON
 import HealthKit
 import SafariServices
+import Alamofire
+import AlamofireImage
 
 
 class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayout, UICollectionViewDelegate, UICollectionViewDataSource, UISearchBarDelegate, SFSafariViewControllerDelegate {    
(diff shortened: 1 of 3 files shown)


BeeSwift/GalleryViewController.swift
```diff
@@ -373,6 +384,21 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
     }
     
     @objc func didFetchGoals() {
+        let thumbs = self.goals.map{ $0.cacheBustingThumbUrl }
+        let graphs = self.goals.map{ $0.cacheBustingGraphUrl }
```
krugerk, line 388: We could certainly debate about the right number here or if this is needed. We download/cache all thumbnails and all graphs (thumbnails first). 
What about slow connections? What about accounts with 70+ goals? Is beeminder server-side checking logs of graphs obtained for analytics to indicate a user navigated to a particular place in the app? 
I guess if we have any doubts, we could leave the graphs downloaded/cached when the user views the goal (GoalVC).

PR #187, opened 2020-09-04
Update bug.md

krugerk:


dreeves:
I'm totally fine with this and am not involved enough in BeemiOS to dictate the format of bug report gissues anyway.

(But in case anyone is curious, I have notes for a followup blog post to https://blog.beeminder.com/bugreports including the following apologia for the Latin names ("reseminata", "expectata", "resultata"):

1. The English phrases aren't consistent. Too many variants of "repro steps" or "what happened", etc.
2. Even if you standardized the words/phrases you wouldn't have greppability.
3. Insufficient gravitas? Not totally serious there but we want to convey "these are the official parts of a Proper Bug Report".

Like if you call it "steps to reproduce" then people think they understand that and describe all the things they did. If you call it "reseminata" they get that that's something technical and specific, which it is. You've solved that nicely with "steps to reproduce the bug from scratch" but that's more unwieldy. But, I should reemphasize, _totally fine_ if you prefer it.

I actually sat down with a linguist friend to figure those terms out because that is my idea of a fun Saturday night. But the fun was coming up with them so I won't be offended if you don't like them!)


PS: In the Beeminder repo we ended up dropping all the boilerplate and have nothing but the following as the BUG.md template:

```
---
name: BUG
about: A Proper Bug Report with repro steps AKA Defectus Demonstrandum
title: ''
labels: BUG
assignees: ''

---

## Reseminata 



## Expectata



## Resultata

```

andrewpbrett:
I don't think it's absurdly cryptic given that there's an English explanation right there; i.e., I doubt anyone who's looked at this template has seen it and not created an issue specifically because of the Latin. Beyond that, I don't have much more of an argument other than I actually *like* that it's not just a boring, character-less form.

I think we err on the side of the status quo. Closing.

PR #207, opened 2020-09-29
Fix #195

### LINKED ISSUE #195: editorial guidelines suggest referring to Apple Health as such
Comments entered with data provided via Apple Health currently are as follows:
 - Automatically entered via iOS Health app
 - Automatically updated via iOS Health app

These both refer to an iOS Health App while the Apple Health [editorial guidelines](https://developer.apple.com/design/human-interface-guidelines/healthkit/overview/) suggest referring to the entity as Apple Health or the Apple Health app.

andrewpbrett:


BeeSwift/JSONGoal.swift
```diff
@@ -633,7 +633,7 @@ class JSONGoal {
             
             if datapoints.count == 0 {
                 let requestId = "\(daystamp)-\(self.minuteStamp())"
-                let params = ["access_token": CurrentUserManager.sharedManager.accessToken!, "urtext": "\(daystamp.suffix(2)) \(datapointValue) \"Automatically entered via iOS Health app\"", "requestid": requestId]
+                let params = ["access_token": CurrentUserManager.sharedManager.accessToken!, "urtext": "\(daystamp.suffix(2)) \(datapointValue) \"Automatically entered via Apple Health app\"", "requestid": requestId]
```
krugerk, line 636: I personally would prefer "via Apple Health" or "via the Apple Health app" (using the and app together).

krugerk: Furthermore, not part of the ticket mentioned but part of the lines adjusted, all other autodata goals I have show 'Auto-entered' and 'Auto-updated'.
I am not sure if it is a beeminder standard or just happened that way because all other autodata goals take place on the backend directly.
How about changing that as well to match the others? 

@dreeves Would you care to standardize on the text for datapoint comments when created and updated using an autodata source?

dreeves: I propose: \"Automatically entered via Apple Health\" -- no "the", no "app"

krugerk: That tackles the 'Apple Health' part.

And what about the automatically entered vs auto-entered (as is the case with autodata goals)?

dreeves: Oh, right, yes, let's get that consistent! Here's an example from my GmailZero goal: "Auto-entered by gmailzero at 17:01. Last count 109 at 09:30:06". Can we model it on that? "Auto-entered by Apple Health at ___". Not sure if there's other useful data like that that could be included in the comment.

krugerk (changes requested):
How about providing another commit with this merge request that covers the 'auto-entered' and 'auto-updated' text as well?

andrewpbrett:
Agree that the time should be separate. Just updated the strings.

krugerk (approved):

PR #224, opened 2020-11-03
Remove duplicated access token in get goal

### LINKED ISSUE #198: Doing a GET on a goal sends the access_token twice
```[tasklist]
### Desiderata
- [ ] Replicate it in current version (6.6)
- [ ] Characterize the severity
```

1. Tap a goal in the gallery to enter the goalVC.
2. Upon `viewDidAppear`, `refreshGoal` will be called which in turn makes a GET request to obtain the goal.
3. It appears the access_token is attached twice to this request:
`https://www.beeminder.com//api/v1/users/<user>/goals/<goalslug>?access_token=<access_token>&datapoints_count=5&access_token=<access_token>`

The access token is appended to the URL directly and RequestManager checks the parameters for the access token but does not check the rest of the URL. Instead it should follow the others, either passing the access token along in the parameters or allowing the default behavior to append the necessary token. 

## Cognata

* https://github.com/beeminder/BeeSwift/blob/d79461c2c685db215d75056d355ba6d6b2dc4fce/BeeSwift/GoalViewController.swift#L573
* https://github.com/beeminder/BeeSwift/blob/d79461c2c685db215d75056d355ba6d6b2dc4fce/BeeSwift/RequestManager.swift#L77

_**Verbata**: tidiness, inconsistencies, redundant parameters,_

krugerk:
fixes #198 by removing the redundant parameter

### DIFF
--- a/BeeSwift/GoalViewController.swift
+++ b/BeeSwift/GoalViewController.swift
@@ -570,7 +570,7 @@ class GoalViewController: UIViewController, UITableViewDelegate, UITableViewData
     }
     
     @objc func refreshGoal() {
-        RequestManager.get(url: "/api/v1/users/\(CurrentUserManager.sharedManager.username!)/goals/\(self.goal.slug)?access_token=\(CurrentUserManager.sharedManager.accessToken!)&datapoints_count=5", parameters: nil, success: { (responseObject) in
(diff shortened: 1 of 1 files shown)


krugerk:
@theospears and this one is very basic

PR #237, opened 2020-12-03
Remove the storage used for the today widget upon signout

### LINKED ISSUE #236: Today widget still populated and allows submitting data even after logout in the app

1. beeminder for ios version 5.8
2. log in via the app
3. have some goals, view the main screen (gallery / list of goals)
4. leave the app, configure today widget
5. note that the widget shows up to three goals
6. return to the app
7. settings / sign out
8. note the app now asks for login data
9. leave the app, return to the screen with the today widget
10. notice the widget still shows the goals of before
11. and allows submitting data (in case at least one of those goals allows manual data entry)

 - no goals listed in today widget after signing out
 - instead of a blank widget, widget should show a message
  - that one is not signed in, no goals configured ... 
  - return to the app, load goals there, ...

List of goals remained available in the widget.
Widget allowed sending data to those goals.

## Extras

 app version: 5.8
 device: iOS device with iOS 14.2

krugerk:
Clears the today widget by removing entries from the storage it uses to find the access token and top three goals.

fixes #236

### DIFF
--- a/BeeSwift/CurrentUserManager.swift
+++ b/BeeSwift/CurrentUserManager.swift
@@ -151,6 +151,10 @@ class CurrentUserManager : NSObject {
         UserDefaults.standard.removeObject(forKey: deadbeatKey)
         UserDefaults.standard.removeObject(forKey: usernameKey)
         UserDefaults.standard.synchronize()
+
+        UserDefaults.standard.removeSuite(named: "group.beeminder.beeminder")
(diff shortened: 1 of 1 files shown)


andrewpbrett:
Is this still being worked on?

krugerk:
No, not actively. 

Listed as draft because there are no tests of the code change.

PR #243, opened 2021-01-07
use an uninterpolated string in eval_gemfile

### LINKED ISSUE #196: github dependabot no longer able to check Gemfile
The dependabot was introduced with #133. It periodically checks github actions and repo gems for updates.

Apparently the dependabot is unable to parse Gemfile containing interpolated strings. Such a string was introduced with Commit 8b4e90dc1260ca34b760f3b3352af38a026021ed when `$ fastlane add_plugin [name] ` updated the Gemfile to evaluate the `fastlane/Pluginfile` as well.

krugerk:
Instead of using the fastlane-generated interpolated string specifiying the plugins_path,
a string is used directly. This allows dependabot to evaluate the Gemfile and thus fixes #196.

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -22,7 +22,8 @@ blocks:
         commands:
           - checkout
           - cache restore
-          - bundle install --path vendor/bundle
+          - bundle config set path 'vendor/bundle'
+          - bundle install
           - cache store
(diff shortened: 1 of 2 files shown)


krugerk:
We can remove the commit in this PR that migrated to the current semaphoreci badges. That change has made its way into master as of PR #250.

PR #251, opened 2021-01-26
Remove health sync notification

### LINKED ISSUE #44: Be smarter about warning me to unlock my phone

andrewpbrett:
Closes #44


BeeSwift/AppDelegate.swift
```diff
@@ -140,19 +137,16 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
     func userNotificationCenter(_ center: UNUserNotificationCenter, willPresent notification: UNNotification, withCompletionHandler completionHandler: @escaping (UNNotificationPresentationOptions) -> Void) {
-        switch notification.request.identifier {
-        case JSONGoal.unlockNotificationIdentifier:
-            // about to present a notification, a reminder to unlock the device
-            // yet the app is active, thus we can abandon the notification
-            completionHandler([])
-        default:
-            completionHandler([.alert, .sound, .badge])
-        }
+        completionHandler([.alert, .sound, .badge])
```
krugerk, line 140: This whole section was introduced in Commit 1b33ac579d3e59bdd and probably could be removed along with the rest of the stuff removed in this PR.

As is now, the presentation options for notifications shown while the app is in the foreground are overwritten. We probably could just use their defaults, which can be achieved by not implementing this method `func userNotificationCenter ... willPresent ...`.

https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/1649518-usernotificationcenter

BeeSwift/AppDelegate.swift
```diff
@@ -140,19 +137,16 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
     }
     
     private func resetStateIfUITesting() {
         if ProcessInfo.processInfo.arguments.contains("UI-Testing") {
             UserDefaults.standard.removePersistentDomain(forName: Bundle.main.bundleIdentifier!)
         }
     }
+    
+    private func removeAllLocalNotifications() {
+        UNUserNotificationCenter.current().removeAllPendingNotificationRequests()
```
krugerk, line 150: hm, possibly this may have been enough for those opening the app regularly. Queued notification requests would have been removed upon app start. 

The lines removed in L143-L150 on the other hand saw to it that the "unlock your phone" message did not appear while the app was in the foreground.

krugerk:
This also renders the entire section **How do I enable iOS reminders?** at https://help.beeminder.com/article/60-ios-app moot. The option is no longer available as of [5.9](https://github.com/beeminder/BeeSwift/releases/tag/5.9).

PR #256, opened 2021-02-03
built in timer getting cancelled

### LINKED ISSUE #5: Built-in timer gets canceled
### Desiderata
- [ ] Confirm this still replicates
- [ ] Theo opines on Andy's idea: put `timingSince` in UserDefaults instead of a variable on the controller
---

## Replicata

1. Start a timer in BeemiOS
2. Task-switch to another app
3. Wait a couple minutes
4. Task-switch back to the timer

## Expectata

That the timer is still running.

## Resultata

It gets canceled. 😭 

## Nota Nebulosa

For comparison, Beedroid's timer is super robust. I think you can even reboot the whole device and that timer will still be running, though I'm not sure I've tried that.

&nbsp;image

## Cognata

* https://github.com/beeminder/BeeSwift/blob/master/BeeSwift/TimerViewController.swift
* #393 
* #290 
* #268 
* #61 
* #17 

_**Verbata**: timer mode, not losing a running timer, baneful bugs,_

krugerk:
Introduce a very basic variant of timer persistence to fix #5.

For those users with 4h long tasks, once the timer is started, the date `timingSince` is persisted, linked to the goal. Once the user returns to the app and eventually the timer interface, the timer is is restored.

Some nice to haves:
 - the timer ticking need not do so while the app is not visible
 - if the timer could be restored, resume the timer as though it had never been stopped
 - UI tests (including sending the app into the background, terminating the app, starting the app, returning to the same goal's timer screen)

### DIFF
--- a/BeeSwift/CurrentUserManager.swift
+++ b/BeeSwift/CurrentUserManager.swift
@@ -33,6 +33,21 @@ class CurrentUserManager : NSObject {
     
     var goals : [JSONGoal] = []
     var goalsFetchedAt : Date = Date()
+
+    func timingSince(for goal: JSONGoal) -> Date? {
+        return UserDefaults.standard.object(forKey: keyTimer(for: goal)) as? Date
+    }
+    func set(timerSince: Date?, for goal: JSONGoal) {
+        let key = keyTimer(for: goal)
+        guard let timingSince = timerSince else {
+            UserDefaults.standard.removeObject(forKey: key)
+            return
+        }
+        UserDefaults.standard.set(timingSince, forKey: key)
+    }
+    private func keyTimer(for goal: JSONGoal) -> String {
+        return "key-timer-\(goal.id)"
+    }
     
     var accessToken :String? {
         return UserDefaults.standard.object(forKey: accessTokenKey) as! String?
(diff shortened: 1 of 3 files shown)


krugerk:
As noted, this would be a very primitive version of this feature.

It would already work for multiple goals. However, it really only handles one particular scenario and fails miserably on others.

It works fine enough for the use case of pauseless timers.
It would produce the wrong result though when pausing the timer.

Let us walk through a few scenarios:

A) timer is being used to record the duration of being in the library. User enters library and after settling, starts the timer at 10:00. The user reads, studies, researches, browses The Guardian, and eventually leaves the building. Just before leaving, upon having finished packing up, the user switches back to the app to stop the timer. This happens at 14:30 (2:30 PM). iOS meanwhile has terminated the app because the app was running a timer in the background needlessly. The app is started and the goals are fetched. The user taps around to enter the library goal and the timer interface from there. The starttime had been saved and is now restored. The timer is seen running and displays the duration since start was tapped. The user sees the value of 4.5 hours and taps to enter that as a datapoint. 

B) user wishes to track time spent replying to emails, actually typing them out, during a session of 'clean the inbox'.
In this scenario the user will often tap stop to pause the counter only to resume it after preparing for the next reply. The user moves from mouse to keyboard then taps start to resume the counter. The counter does not show the time between when it was originally started but instead the sum of all of the durations the timer has been running. The screen blackens while the user is typing out a long reply and when returning to the smartphone for the app, the user needs to enter credentials again to pass the Lock Screen. Returning to the app/goal/timer, the user sees a big number now as the value of the timer. It is clearly too high. The user has the feeling at least half the time is being allocated to sorting and thinking up replies and only a small amount should be under typing them up.

PR #273, opened 2021-03-30
Shortcuts

andrewpbrett:


Podfile
```diff
@@ -31,6 +35,8 @@ post_install do |installer|
   installer.pods_project.targets.each do |target|
     target.build_configurations.each do |config|
       config.build_settings['SWIFT_VERSION'] = '4.0'
```
krugerk, line 37: pods are set to use Swift 4, still

BeeSwiftIntents/AddDataIntentHandler.swift
```diff
@@ -0,0 +1,40 @@
...
+    }
+    
+    func confirm(intent: AddDataIntent,
+               completion: @escaping (AddDataIntentResponse) -> Void) {
+        completion(AddDataIntentResponse(code: .ready, userActivity: nil))
+    }
+
+    func handle(intent: AddDataIntent,
+              completion: @escaping (AddDataIntentResponse) -> Void) {
+        let params = ["urtext": "^ 1", "requestid": UUID().uuidString]
```
krugerk, line 32: `params` seems to be ignored

BeeSwiftIntents/AddDataIntentHandler.swift
```diff
@@ -0,0 +1,40 @@
...
+        
+        RequestManager.addDatapoint(urtext: "^ 1", slug: intent.goal!) { (response) in
```
krugerk, line 34: It would make sense to write this without ever needing to force unwrap, seen here with `intent.goal!`

BeeSwiftIntents/AddDataIntentHandler.swift
```diff
@@ -0,0 +1,40 @@
...
+//  Created by Andrew Brett on 3/28/21.
+//  Copyright © 2021 APB. All rights reserved.
+//
+
+import Foundation
+import Intents
+
+class AddDataIntentHandler: INExtension, AddDataIntentHandling {
+    func resolveValue(for intent: AddDataIntent, with completion: @escaping (AddDataValueResolutionResult) -> Void) {
+        print("foo")
```
krugerk, line 14: The use of a logging framework along with helpful logging messages, for use during development, debugging, and troubleshooting sure do make sense.

BeeSwift/RequestManager.swift
```diff
@@ -84,4 +84,9 @@ class RequestManager {
         return params
     }
 
+    class func addDatapoint(urtext: String, slug: String, success: ((Any?) -> Void)?, errorHandler: ((Error?, String?) -> Void)?) {
+        let params = ["urtext": urtext, "requestid": UUID().uuidString]
```
krugerk, line 88: I am not sure the benefit of sending the requestId when it is just some random string not being tracked by the app.

BeeSwift/GoalViewController.swift
```diff
@@ -546,15 +556,18 @@ class GoalViewController: UIViewController, UITableViewDelegate, UITableViewData
         hud?.mode = .indeterminate
         self.submitButton.isUserInteractionEnabled = false
         self.scrollView.scrollRectToVisible(CGRect(x: 0, y: 0, width: 0, height: 0), animated: true)
-        let params = ["urtext": self.urtextFromTextFields(), "requestid": UUID().uuidString]
-        
-        RequestManager.post(url: "api/v1/users/\(CurrentUserManager.sharedManager.username!)/goals/\(self.goal.slug)/datapoints.json", parameters: params, success: { (responseObject) in
+        if #available(iOS 12.0, *) {
+            self.donateAddDataIntent(value: self.valueTextField.text!)
+        } else {
```
krugerk, line 561: these `else`s should be removed as they are empty and add no value.

BeeSwift/GoalViewController.swift
```diff
@@ -546,15 +556,18 @@ class GoalViewController: UIViewController, UITableViewDelegate, UITableViewData
...
+            // do nothing
+        }
+        RequestManager.addDatapoint(urtext: self.urtextFromTextFields(), slug: self.goal.slug) { (responseObject) in
             self.commentTextField.text = ""
             self.refreshGoal()
             self.pollUntilGraphUpdates()
             self.submitButton.isUserInteractionEnabled = true
             CurrentUserManager.sharedManager.fetchGoals(success: nil, error: nil)
-        }) { (error, errorMessage) in
+        } errorHandler: { (error, errorMessage) in
```
krugerk, line 570: yay!

BeeSwift/GoalViewController.swift
```diff
@@ -41,6 +43,14 @@ class GoalViewController: UIViewController, UITableViewDelegate, UITableViewData
         }
         self.title = self.goal.slug
         
+        if #available(iOS 12.0, *) {
+            let viewGoalActivity = self.viewGoalShortcut()
+            self.userActivity = viewGoalActivity
```
krugerk, line 48: Supporting iOS Search via NSUserActivity has been around since iOS 8. It looks like this is meant for use inside the Shortcuts app.


Beeminder "add data" available within Shortcuts:
Simulator Screen Shot - iPhone 12 Pro Max - 2021-03-30 at 10 08 53
Simulator Screen Shot - iPhone 12 Pro Max - 2021-03-30 at 10 09 04

krugerk: Having the data pop up in search is not that far off. 

--




some screenshots from using NSUserActivity to integrate with iOS Search, from Aug 2020:
Simulator Screen Shot - iPhone SE (2nd generation) - 2020-08-06 at 17 25 12

Simulator Screen Shot - iPhone SE (2nd generation) - 2020-08-06 at 14 31 25

BeeSwift/ConfigureNotificationsViewController.swift
```diff
@@ -121,7 +121,7 @@ extension ConfigureNotificationsViewController : UITableViewDataSource, UITableV
     }
     
     func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-        let cell = tableView.dequeueReusableCell(withIdentifier: self.cellReuseIdentifier) as! SettingsTableViewCell!
+        let cell = tableView.dequeueReusableCell(withIdentifier: self.cellReuseIdentifier) as! SettingsTableViewCell?
```
krugerk, line 124: The combination of as**!** and SettingsTableViewCell**?** seems to be contradictory. 

On the one hand, sure maybe we can assume these do not break. On the other, how about

```

guard
  let cell = tableView.dequeueReusableCell(withIdentifier: self.cellReuseIdentifier) as? SettingsTableViewCell
else
  { return UITableViewCell() }

```

BeeSwift/AppDelegate.swift
```diff
@@ -149,4 +149,25 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
     private func removeAllLocalNotifications() {
         UNUserNotificationCenter.current().removeAllPendingNotificationRequests()
     }
+    func application(_ application: UIApplication, continue userActivity: NSUserActivity, restorationHandler: @escaping ([UIUserActivityRestoring]?) -> Void) -> Bool {
+        if #available(iOS 12.0, *) {
+            if let intent = userActivity.interaction?.intent as? AddDataIntent {
+                guard let value = intent.value, let slug = intent.goal else { return false }
+                RequestManager.addDatapoint(urtext: "^ \(value)", slug: slug) { (response) in
```
krugerk, line 156: Where can documentation be found explaining this shorthand here?

setting
urText to "^ value" ?



(^ is translated on the server to mean "today" and value is the value being added as part of the datapoint, right? And one can write directly to urText rather than specifying daystamp and value individually, as outlined in the [api docs](https://api.beeminder.com/#postdata).)

krugerk:
- It would be nice if the 'add data to goal' provided a list of goals rather than requiring the user to type in the goal's slug. (Both options exist; the add data to goal <slug> should be given priority (listed first).
- Upon signing out, the 'view Beeminder goal with slug' entries for Shortcuts are not removed.
- After having defined a shortcut to add data using the new BeeSwift support for iOS Shortcuts app, when running the shortcut without adequate Internet access (offline, no connection to Beeminder.com's api endpoint), the operation eventually times out. image

PR #275, opened 2021-04-04
avoid keyboard overlapping gallery cells

### LINKED ISSUE #200: While searching last goal is covered by keyboard
A user has reported this

 > Just a quick, minor bug report re. the search bar:
If you have enough results of a search that you have to scroll, it’s impossible to select the last one on the list, because it’s tucked behind the keyboard and can’t be reached. You can keep scrolling to see it, but can’t tap on it.

Refining the search as a workaround works but is inconvenient.

 > Yes, though it’s not ideal, friction-wise.
We can get to any goal we want by typing more of the single goal’s slug but, for instance, sometimes I want to check all of my ios-[…] goals or all of my mh-[…] goals and the friction is fairly palpable (admittedly that’s in part b/c of the dash being on a different keyboard which requires that, to narrow it down further so that I can get to the last one on the list, I switch keyboards, hit the dash, switch keyboards, and keep tying. Not the end of the world, of course but palpable, friction-wise.

So, the task? 
The last goal (cell) should be fully visible.

## reproduce
It is enough to bring up the search area with a list of goals that do not fit all at one time on the screen.
Even without entering anything in the search box, one can no longer scroll such that the bottom most goal is completely visible.

App: 5.7
Device: ?

krugerk:
adjusts the gallery's insets to account for the keyboard

fixes #200

### DIFF
--- a/BeeSwift/GalleryViewController.swift
+++ b/BeeSwift/GalleryViewController.swift
@@ -35,12 +35,16 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
     override func viewDidLoad() {
         super.viewDidLoad()
         
-        NotificationCenter.default.addObserver(self, selector: #selector(self.handleSignIn), name: NSNotification.Name(rawValue: CurrentUserManager.signedInNotificationName), object: nil)
-        NotificationCenter.default.addObserver(self, selector: #selector(self.handleSignOut), name: NSNotification.Name(rawValue: CurrentUserManager.signedOutNotificationName), object: nil)
-        NotificationCenter.default.addObserver(self, selector: #selector(self.openGoalFromNotification(_:)), name: NSNotification.Name(rawValue: "openGoal"), object: nil)
-        NotificationCenter.default.addObserver(self, selector: #selector(self.handleCreateGoalButtonPressed), name: NSNotification.Name(rawValue: "createGoalButtonPressed"), object: nil)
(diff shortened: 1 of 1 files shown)


krugerk:
#200 mentioned that this was an issue while searching and that is true but it would have been an issue anytime the gallery contained enough cells and the keyboard was shown: where the gallery has a full screen worth of goals (and new goal) cells, the last two are covered by the keyboard.

BeeSwift/GalleryViewController.swift
```diff
@@ -236,6 +240,33 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
...
+        guard notification.name == UIResponder.keyboardWillHideNotification else { return }
+
+        collectionView.contentInset = .zero
+        collectionView.scrollIndicatorInsets = .zero
+    }
+
+    @objc func keyboardFrameWillChange(notification: Notification) {
+        guard let collectionView = self.collectionView else { return }
+        guard let keyboardValue = notification.userInfo?[UIResponder.keyboardFrameEndUserInfoKey] as? NSValue else { return }
+        guard notification.name != UIResponder.keyboardWillHideNotification else { return }
```
krugerk, line 254: suppose this guard can appear a line earlier

krugerk: or how about checking for the particular name expected instead?

krugerk:
@theospears might you review this bugfix merge request?

krugerk:
A new approach, available with iOS 15, was proposed in #536.

Issue #293, opened 2022-08-27
Cannot add data through shortcuts integration


theospears:
# Defectus Demonstrandum

As reported on the forum: https://forum.beeminder.com/t/ios-shortcut-integration/10024

Attempting to run a shortcut which adds data to beeminder produces the error: “Could not run Add Data - There was a problem communicating with the app”.

# Nota Nebulosa

iOS Logs show:
```
[__extensionPrincipalClass conformsToProtocol:@protocol(NSExtensionRequestHandling)] - /Library/Caches/com.apple.xbs/Sources/ExtensionFoundation/ExtensionFoundation/Source/NSExtension/NSExtensionSupport/EXConcreteExtensionContextVendor.m:109: NSExtensionPrincipalClass does not conform to NSExtensionRequestHandling protocol!
```

This looks plausibly related.

PR #300, opened 2022-09-04
Share all UserDefaults with the app group

theospears:
The UserDefaults system has a number of different namespaces. BeeSwift by default has been using one which is not shared with extensions, and done some extra bookkeeping to share particular values with the today widget. This is not sufficient for Intents, and also adds complexity, so instead migrate to storing all settings in a location accessible to all apps in the app group.

To support branch switching and downgrading, at present all values are migrated and double-written to both locations. Once we are confident all users have migrated we can remove this migration code.

This is a prerequisite for #293

Test Plan:
* [ ] Log into an older version of the app, upgrade, and check we are still logged in
* [ ] Log out on new version, downgrade, and check we are logged out
* [ ] Log in on new version, downgrade, check we are still logged in
* [ ] Confirm the today widget still works

### DIFF
--- a/BeeSwift/CurrentUserManager.swift
+++ b/BeeSwift/CurrentUserManager.swift
@@ -21,6 +21,8 @@ class CurrentUserManager : NSObject {
     static let goalsFetchedNotificationName = "com.beeminder.goalsFetchedNotification"
     static let healthKitMetricRemovedNotificationName = "com.beeminder.healthKitMetricRemovedNotification"
     
+    fileprivate let beemiosSecret = "C0QBFPWqDykIgE6RyQ2OJJDxGxGXuVA2CNqcJM185oOOl4EQTjmpiKgcwjki"
+    
     fileprivate let accessTokenKey = "access_token"
     fileprivate let usernameKey = "username"
     fileprivate let deadbeatKey = "deadbeat"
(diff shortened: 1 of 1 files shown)


BeeSwift/CurrentUserManager.swift
```diff
@@ -21,79 +21,119 @@ class CurrentUserManager : NSObject {
...
     fileprivate let usernameKey = "username"
     fileprivate let deadbeatKey = "deadbeat"
     fileprivate let defaultLeadtimeKey = "default_leadtime"
     fileprivate let defaultAlertstartKey = "default_alertstart"
     fileprivate let defaultDeadlineKey = "default_deadline"
     fileprivate let beemTZKey = "timezone"
-    fileprivate let beemiosSecret = "C0QBFPWqDykIgE6RyQ2OJJDxGxGXuVA2CNqcJM185oOOl4EQTjmpiKgcwjki"
+    
+    var allKeys: [String] {
+        return [accessTokenKey, usernameKey, deadbeatKey, defaultLeadtimeKey, defaultAlertstartKey, defaultDeadlineKey, beemTZKey]
```
krugerk, line 35:
```suggestion
   [accessTokenKey, usernameKey, ...]
```


Return in this context is redundant and can be removed

BeeSwift/CurrentUserManager.swift
```diff
@@ -21,79 +21,119 @@ class CurrentUserManager : NSObject {
...
+    }
+    
+    /// Migrate settings values from the standard store to a group store
+    ///
+    /// Originally BeeSwift stored all configuration values in the standard UserDefaults store. However
+    /// these values are not available within extensions. To address this now values are stored in a
+    /// group-scoped settings object. Values written by old versions of the app may be in the previous store
+    /// so we migrate any such values on initialization.
+    private func migrateValues() {
+        for key in allKeys {
```
krugerk, line 52: How about

```suggestion
   allKeys.forEach ...
```

BeeSwift/CurrentUserManager.swift
```diff
@@ -21,79 +21,119 @@ class CurrentUserManager : NSObject {
...
+            let standardValue = UserDefaults.standard.object(forKey: key)
+            let groupValue = userDefaults.object(forKey: key)
+            
+            if groupValue == nil && standardValue != nil {
```
krugerk, line 56: A `guard` might look better here

krugerk (approved):
Nice

BeeSwift/CurrentUserManager.swift
krugerk: SwiftFormat could remove this for us: https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#redundantreturn

BeeSwift/CurrentUserManager.swift
theospears: I agree this is also correct. Are there reasons to choose one form over the other? (Googling didn't turn up anything compelling.) The rest of the codebase seems to mainly, but not entirely, use the `.forEach` form.

krugerk: In this case I prefer the latter since appears to match more closely the functional programming style.

theospears:
Thanks for the feedback! I'm going to consider it for future PRs and cleanup, but not going to block merging on it for now.

PR #301, opened 2022-09-04
Fix all compiler warnings

theospears:
This addresses the compiler warnings generated while building BeeSwift.

The majority of these are deprecation warnings. The code is updated according to the advice in the warnings. There were also a few changes to building and linking.

Test Plan:
On a single device:
* [ ] Refresh goal while offline and check error
* [ ] Check reset password button still works and brings up web view
* [ ] End a timer session while offline and check error

On devices with and without a notch:
* [ ] Verify today widget still renders and accepts input
* [ ] Check that healthkit metric chooser looks reasonable
* [ ] Check that healthkit metric removal looks reasonable
* [ ] Check that edit data point UI looks reasonable
* [ ] Check that goal notification settings looks reasonable
* [ ] Check that gallery view looks reasonable

### DIFF
--- a/BeeSwift/GoalViewController.swift
+++ b/BeeSwift/GoalViewController.swift
@@ -552,7 +552,6 @@ class GoalViewController: UIViewController, UITableViewDelegate, UITableViewData
             let alertController = UIAlertController(title: "Error", message: "Failed to add datapoint", preferredStyle: .alert)
             alertController.addAction(UIAlertAction(title: "OK", style: .cancel))
             self.present(alertController, animated: true)
-            //UIAlertView(title: "Error", message: "Failed to add datapoint", delegate: nil, cancelButtonTitle: "OK").show()
         }
     }
     
--- a/BeeSwift/GoalViewController.swift
+++ b/BeeSwift/GoalViewController.swift
@@ -549,7 +549,10 @@ class GoalViewController: UIViewController, UITableViewDelegate, UITableViewData
         } errorHandler: { (error, errorMessage) in
             self.submitButton.isUserInteractionEnabled = true
             MBProgressHUD.hideAllHUDs(for: self.view, animated: true)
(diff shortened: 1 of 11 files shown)


BeeSwift/GoalViewController.swift
```diff
@@ -549,7 +549,10 @@ class GoalViewController: UIViewController, UITableViewDelegate, UITableViewData
         } errorHandler: { (error, errorMessage) in
             self.submitButton.isUserInteractionEnabled = true
             MBProgressHUD.hideAllHUDs(for: self.view, animated: true)
-            UIAlertView(title: "Error", message: "Failed to add datapoint", delegate: nil, cancelButtonTitle: "OK").show()
+            let alertController = UIAlertController(title: "Error", message: "Failed to add datapoint", preferredStyle: .alert)
+            alertController.addAction(UIAlertAction(title: "OK", style: .cancel))
+            self.present(alertController, animated: true)
+            //UIAlertView(title: "Error", message: "Failed to add datapoint", delegate: nil, cancelButtonTitle: "OK").show()
```
krugerk, line 555: No need to commit commented out code

krugerk (approved):
Nice

cantastic:
1. I turned off wifi and cell data on my Iphone, then refreshed a goal in the ios app, and this appeared:
"Eror could not refresh graph". my only option was "OK".

krugerk:
> I turned off wifi and cell data on my Iphone, then refreshed a goal in the ios app, and this appeared:
"Eror could not refresh graph". my only option was "OK".


Ah, yes. The app as is does not support an offline mode. And this Merge Request deals mostly with tackling some 'technical debt'. Previously an alert dialog was shown in the aforementioned scenario. Code was updated to present this dialog in a more modern way. Expected behavior is that the dialog is still shown like it was before (and that that segment of the code compiles without the 'this style of alert dialog is deprecated' warning).

Thanks for engagement!

PR #303, opened 2022-09-05
Fix the AddData Intent

### LINKED ISSUE #293: Cannot add data through shortcuts integration

theospears:
Previous commits had begun to implement an intent to add data points to beeminder. This follows up on that work, adding the necessary extension routing and parameter checking to make intents support work without launching the app. This should be seen as a minimum viable implementation, there remain many possible improvements to e.g. error handling, and suggesting allowed goal names.

As datapoint submission is now handled the the extension this removes the code which submitted them as part of app startup.

Fixes #293

Test Plan:
* [X] Created a shortcut which added a data point
* [X] Ran this shortcut, observed it finished successfully and that the data point was added
* [X] Tried to run without a goal name set, or with an invalid goal name, and observed a sensible error.

### DIFF
--- a/BeeSwift/AddDataIntentHandler.swift
+++ b/BeeSwift/AddDataIntentHandler.swift
@@ -30,17 +30,28 @@ class AddDataIntentHandler: NSObject, AddDataIntentHandling {
     }
     
     func confirm(intent: AddDataIntent) async -> AddDataIntentResponse {
-        AddDataIntentResponse(code: .ready, userActivity: nil)
+        if intent.goal != nil && intent.value != nil {
+            return AddDataIntentResponse(code: .ready, userActivity: nil)
+        } else {
+            return AddDataIntentResponse(code: .failure, userActivity: nil)
+        }
     }
 
     func handle(intent: AddDataIntent,
               completion: @escaping (AddDataIntentResponse) -> Void) {
-        guard let datapointValue = intent.value else { return }
-        
-        RequestManager.addDatapoint(urtext: "^ \(datapointValue)", slug: intent.goal!) { (response) in
-            completion(AddDataIntentResponse.success(goal: intent.goal!))
+        guard let goal = intent.goal else {
+            completion(AddDataIntentResponse.failure(goal: ""))
+            return
+        }
+        guard let value = intent.value else {
+            completion(AddDataIntentResponse.failure(goal: goal))
+            return
+        }
+                
+        RequestManager.addDatapoint(urtext: "^ \(value)", slug: goal) { (response) in
+            completion(AddDataIntentResponse.success(goal: goal))
         } errorHandler: { (error, errorMessage) in
-            completion(AddDataIntentResponse.failure(goal: intent.goal!))
+            completion(AddDataIntentResponse.failure(goal: goal))
         }
     }
 }
--- a/BeeSwift/AddDataIntentHandler.swift
+++ b/BeeSwift/AddDataIntentHandler.swift
@@ -21,9 +21,9 @@ class AddDataIntentHandler: NSObject, AddDataIntentHandling {
     }
     
     func resolveGoal(for intent: AddDataIntent) async -> INStringResolutionResult {
-        if let goal = intent.goal {
+        if let goalSlug = intent.goal {
             // TODO: We should validate this is a valid slug
-            return INStringResolutionResult.success(with: goal)
+            return INStringResolutionResult.success(with: goalSlug)
         } else {
             return INStringResolutionResult.needsValue()
         }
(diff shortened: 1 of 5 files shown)


BeeSwift/AddDataIntentHandler.swift
```diff
@@ -12,28 +12,46 @@ import BeeKit
...
+        } else {
+            return INStringResolutionResult.needsValue()
+        }
     }
     
-    func confirm(intent: AddDataIntent,
-               completion: @escaping (AddDataIntentResponse) -> Void) {
-        completion(AddDataIntentResponse(code: .ready, userActivity: nil))
+    func confirm(intent: AddDataIntent) async -> AddDataIntentResponse {
+        if intent.goal != nil && intent.value != nil {
```
krugerk, line 33: How about an extension with computed property `isValid`?

BeeSwift/AppDelegate.swift
```diff
@@ -153,22 +153,16 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
     func application(_ application: UIApplication, continue userActivity: NSUserActivity, restorationHandler: @escaping ([UIUserActivityRestoring]?) -> Void) -> Bool {
         if #available(iOS 12.0, *) {
```
krugerk, line 154: While here, how about turning this line into a `guard`?

BeeSwiftIntents/IntentHandler.swift
```diff
@@ -0,0 +1,17 @@
+//
+//  IntentHandler.swift
+//  BeeSwiftIntents
+//
+//  Created by Theo Spears on 9/1/22.
+//  Copyright © 2022 APB. All rights reserved.
```
krugerk, line 6: Is this still accurate? Code you type is Copyright _andrewpbrett_ ?

BeeSwiftIntents/IntentHandler.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+
+import Foundation
+import Intents
+
+@available(iOS 14.0, *)
+class IntentHandler : INExtension {
+    override func handler(for intent: INIntent) -> Any? {
+        return AddDataIntentHandler()
```
krugerk, line 15: [redundantReturn](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#redundantreturn)

BeeSwift/AddDataIntentHandler.swift
```diff
@@ -12,28 +12,46 @@ import BeeKit
...
+        }
     }
 
     func handle(intent: AddDataIntent,
               completion: @escaping (AddDataIntentResponse) -> Void) {
-        guard let datapointValue = intent.value else { return }
-        
-        RequestManager.addDatapoint(urtext: "^ \(datapointValue)", slug: intent.goal!) { (response) in
-            completion(AddDataIntentResponse.success(goal: intent.goal!))
+        guard let goal = intent.goal else {
```
krugerk, line 42: How about naming this variable `goalSlug` or `slug`? It is not a full JSONGoal but merely a String containing the goal's slug.

BeeSwiftIntents/IntentHandler.swift
theospears: I think my preference would be to strip all of these comment headers - copyright on opensource files with multiple contributors ends up pretty complex, and all the other info in the headers is redundant.

BeeSwift/AddDataIntentHandler.swift
theospears: This seems reasonable. In a perfect world we would rename the property on the Intent to `goalSlug` and then do `guard let goalSlug = intent.goalSlug ...`. However, given the intent already exists, renaming the property would break existing shortcuts which I'm reluctant to do.

This means the best we can get to is `guard let goalSlug = intent.goal ...` - trading off a better name for less consistency. So not a costless win, but maybe still net better?

BeeSwift/AppDelegate.swift
theospears: I don't think this behaves quite like a guard, because we still want to inspect `userActivity.userInfo?` later even if it's not satisfied?

krugerk: Of course the function will need to return something when running on lower than ios 12. The guard's else provides for this.

The meat of the function though happens inside that if and using a `guard` here should allow one to compose much more legible code.

See https://github.com/raywenderlich/swift-style-guide#golden-path

BeeSwift/AddDataIntentHandler.swift
krugerk: Here I only meant the new code which introduced the `guard let goal = ...` variable. That local variable might better named slug.

PR #316, opened 2022-10-08
Run tests through github actions on PR

theospears:
This project was previously configured to use semaphore for unit tests. However lack of credits meant a slow queue. Github actions now has the capability to run these tests, and is more consistent with what is used elsewhere at beeminder.

This causes github to run the test suite on PRs.

Note: The test suite is currently broken. That is not addressed by this PR.


theospears:
This is very similar to #315 but not created from a fork, thus should get some additional test result UI which github permissions don't allow in forks.

PR #323, opened 2022-10-16
Fix display of & in datapoint list

### LINKED ISSUE #283: allow & and other characters in datapoint comments
A user of the ios app reports that datapoint comments including `&` character don't show the ampersand correctly. We're escaping it or something.

(TODO: I requested a screenshot.)

theospears:
For some reason the canonical representation of data points html-encodes `&`. To avoid this we instead reconstruct the text from the individual fields. In the future we could go further and use columns, but for now we just re-assemble the original text output.

Fixes #283

Test Plan:
Added & to a datapoint and viewed the goal in the simulator


psik-sh:
Thanks @theospears!

PR #371, opened 2023-02-11
Add additional logging to understand background update behavior

theospears:
The beeminder app is meant to do quite a lot of work in the background, from updating goals
to also refreshing the badge count. This doesn't appear to always reliably fire, so we add
some logging to have more visibility into what is and is not happening.

Testing:
Ran on device, observed logs firing

PR #372, opened 2023-02-11
Avoid no-op datapoint updates

theospears:
Previously we would always send new data point values to beeminder, even if the
old and new values were the same. This could lead to a large number of unnecessary
API calls, slowing things down and increasing server load. Now instead we check if
the existing value matches (with a small allowable epsilon) and only update on differences

Testing:
Ran the code on device and observed fewer (but some) updates

### DIFF
--- a/BeeSwift/Goal.swift
+++ b/BeeSwift/Goal.swift
@@ -17,7 +17,7 @@ class Goal {
 
     // Ignore automatic datapoint updates where the difference is a smaller fraction than this. This
     // prevents effectively no-op updates due to float rounding
-    private let datapointValueEpsilon = 0.00001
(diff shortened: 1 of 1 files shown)


theospears:
* master
  * **PR #371** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/371" target="_blank">Graphite</a> 
    * **PR #372** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/372" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/beeminder/BeeSwift/372?utm_source=stack-comment).

PR #373, opened 2023-02-11
Rate limit observer-triggered apple health updates

theospears:
For some metrics (principally active calories) we notice apple calling
our observer query a very large number of times in short succession. This
results in us doing duplicated work. Avoid this by rate limiting how often
we will do an update. This will not affect manual updates by pressing the
sync button.

Note that we still debounce even if an update fails. This seems more or less
reasonable - if an update fails we have no particular reason to believe a retry
will succeed, and we'll get it on the next update.

Testing:
Ran on device and observed logs showing multiple updates are skipped.


theospears:
* master
  * **PR #373** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/373" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/beeminder/BeeSwift/373?utm_source=stack-comment).

PR #381, opened 2023-03-02
Maintain order of goals

theospears:
Switching to storing goals in a dictionary for quick lookup meant we were no longer
deterministically maintaining order, leading to some shuffle when goals were refreshed.
Avoid this by using an OrderedDictionary so the order from the server is maintained.

Testing:
Ran in simulator. Observed pulling to refresh kept the same order


theospears:
* master
  * **PR #381** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/381" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/beeminder/BeeSwift/381?utm_source=stack-comment).

PR #394, opened 2023-07-16
Store beeminder access token in iOS keychain

theospears:
This moves the storage of the beeminder access key from UserDefaults
to the iOS keychain. As this is a secret that can be used to impersonate
the user, this is a more appropriate storage location.

We use the most lax security rules. In particular this means:
* The token is available as soon as the phone is first unlocked on boot,
  even if it is then re-locked.
* The token will be transfered to a new device on migration.

This seems appropriate, as the app may want to be able to perform background
updates, and migrating to a new device is convenient. Not having access prior
to first unlock should be fine, as health data and shortcut automations are
both unavailable before this happens.

Note because this deletes the insecure record, moving back to an older version
after this change will require logging in again.

Testing:
* Wrote unit tests for retrieval and migration
* Ran on simulator and verified upgrading didn't require logging in again


theospears:
* master
  * **PR #394** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/394" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/beeminder/BeeSwift/394?utm_source=stack-comment).

PR #395, opened 2023-07-16
Access token is always supplied by the RequestManager

theospears:
The BeeSwift RequestManager adds authentication information to all
requests. However, in the past this would not work in some scenarios, for
example when code is running from extensions. To work around this some actions
would also explicitly add the access token to the URL.

RequestManager should now always add the credentials (as they are stored in a
location shared with the app group) and so this should no longer be necessary.
Remove all locations in the code base where we manually passed through credentials
to simplify the code, and reduce the chance of them ending up in logs.

Testing:
* [x] Verify you can still add and refresh in the today extension
* [x] Verify you can still add data through shortcuts
* [x] Verify modified endpoints in the main beeswift app still work


theospears:
* master
  * **PR #394** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/394" target="_blank">Graphite</a> 
    * **PR #395** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/395" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/beeminder/BeeSwift/395?utm_source=stack-comment).

PR #398, opened 2023-07-31
Do not crash today widget when goals unavailable

### LINKED ISSUE #236: Today widget still populated and allows submitting data even after logout in the app

theospears:
The today widget previously assumed that the app had made goals available, and
would crash if it was empty. This could happen if the user had never logged in,
or never fetched goals. Now instead just show an empty UI. We could ideally show
an error state, but given today widgets are deprecated we'll accept not crashing.

To test this more easily also clear the today widget on logout. This should fix
#236.

Testing:
Logged out and verified the today widget did not crash in the xcode debugger

### DIFF
--- a/BeeSwift/GoalManager.swift
+++ b/BeeSwift/GoalManager.swift
@@ -198,6 +198,10 @@ actor GoalManager {
         self.goalsBox.set(nil)
         self.goalsFetchedAt = Date(timeIntervalSince1970: 0)
         currentUserManager.removeObject(forKey: cachedLastFetchedGoalsKey)
+
+        if let sharedDefaults = UserDefaults(suiteName: Constants.appGroupIdentifier) {
+            sharedDefaults.removeObject(forKey: "todayGoalDictionaries")
+        }
     }
 
     // MARK: Today Widget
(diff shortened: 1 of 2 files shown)


theospears:
* master
  * **PR #398** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/398" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/beeminder/BeeSwift/398?utm_source=stack-comment).

PR #399, opened 2023-07-31
Do not crash if goals do not have a mathishard property

theospears:
While it is intended that the api returns a value for `mathishard` for all
goals there appear to exist some corrupt goals where this is not the case, and
this currently causes the app to crash, preventing access to all the other
goals. Be more tolerant to this property not existing.

Testing:
Wrote unit tests to validate the app can parse these response payloads without
crashing.


theospears:
* master
  * **PR #399** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/399" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/beeminder/BeeSwift/399?utm_source=stack-comment).

PR #400, opened 2023-07-31
Fix crash when modifying notification settings

theospears:
In #344 we made various functions run their behavior inside async Tasks. In
the case of timer callback methods this was unsafe as it meant we ran the code
after the method returned, which means properties on the timer were no longer
set as expected. Solve this by moving all access to timer properties outside
the Task.

Testing:
Repeatedly modified settings in the app without exercising the crash


theospears:
* master
  * **PR #400** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/400" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/beeminder/BeeSwift/400?utm_source=stack-comment).

PR #404, opened 2023-09-02
Use consistent colors throughout the app

theospears:
This adds all the colors used through the app to the beeminder UIColor
extension and updates call sites to use it. This should make it possible
to update colors in one place if we want to make future tweaks.

This only updates colors which are controlled by the app, so for example
will not affect the charts, or the borders around them.

Testing:
Loaded the app and clicked around


theospears:
* master
  * **PR #404** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/404" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/beeminder/BeeSwift/404?utm_source=stack-comment).

PR #408, opened 2023-09-14
All targets should codegen intents

theospears:
We are also seeing some build failures due to issues with missing symbols which
should be generated from intents. It appears that we are not running codegen
for intents for some targets, which may be contributing. Switch so all projects
run codegen.


theospears:
* master
  * **PR #408** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/408" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/beeminder/BeeSwift/408?utm_source=stack-comment).

PR #414, opened 2023-10-03
Fix race condition crash when getting goal connections

theospears:
There is a mutable map linking goals to health kit data sources. While all writes to
this were protected by a mutex, reads were not. It appears it is not safe to read from
this connection while writing to it, so this lead to crashes. Fix by also wrapping the
read.

Testing:
Launched the app and checked it could sync goals


theospears:
* master
  * **PR #414** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/414" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/beeminder/BeeSwift/414?utm_source=stack-comment).

PR #419, opened 2023-11-20
Read and write requestids for datapoints with HealthKit

theospears:
This prepares for supporting point-by-point syncing of data from HealthKit. In particular
1) HealthKit data sources now provide a requestid for the datapoints they generate
2) We always read requestid from the server (now they are always sent)

This will start writing new IDs immediately, and in the future means we can use the IDs to sync when we have multiple datapoints per day.

Testing:
* [ ] Deleted some existing data points, synced with apple heath, and checked they got requestids as expected
* [ ] Checked the app can still update existing points via apple health
* [ ] Confirmed data points still load without crashing

### DIFF
--- a/BeeSwift/DataPoint.swift
+++ b/BeeSwift/DataPoint.swift
@@ -4,6 +4,7 @@ import Foundation
 import SwiftyJSON
 
 protocol DataPoint {
+    var requestid: String { get }
     var daystamp: String { get }
     var value: NSNumber { get }
     var comment: String { get }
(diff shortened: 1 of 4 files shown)


theospears:
* master
  * **PR #419** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/419?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/beeminder/BeeSwift/419?utm_source=stack-comment).

PR #437, opened 2024-01-28
Correctly clear Apple Health Integration

### LINKED ISSUE #248: Can't set deadline on a goal that used to be an Apple Health goal
```[tasklist]
### Desiderata
- [ ] Bee opines on whether this is still a bug
```

## Replicata

1. Have a goal
2. Use the Beeminder app (tested with v5.8) to link that goal with any of the Apple Health metrics
3. Allow or deny Beeminder the app access to the relevant Health metric
4. Use the Beeminder app to remove the link between the goal and the Health metric
5. Go to the website and try to change the goal's deadline

## Expectata

That it lets me change the deadline.

## Resultata

I can't.

## Nota Nebulosa

Beeminder iOS app sets the autodata source to an empty string. This non-nil results in the deadline not being adjustable (on the website).

Forum post about this issue, from Feb 2018: 
https://forum.beeminder.com/t/accidentally-connected-my-goal-to-apple-health/3781/6

_**Verbata**: apple health, custom deadlines, adjusting deadlines, ii object for autadata, setting autodata source,_

theospears:
Previously when attempting to disconnect Apple Health goals they would enter a state where they no longer updated, but beeminder.com still considered them autodata, and thus e.g. changing the deadline was forbidden. This was because we were setting the autodata name to an empty string, rather than null. Sending null was not supported via the URL encoding schema the app was using, so switch to sending parameters as JSON instead and send nil.

Fixes #248

### DIFF
--- a/BeeKit/Managers/RequestManager.swift
+++ b/BeeKit/Managers/RequestManager.swift
@@ -30,7 +30,8 @@ public class RequestManager {
     private let logger = Logger(subsystem: "com.beeminder.beeminder", category: "RequestManager")
     
     func rawRequest(url: String, method: HTTPMethod, parameters: [String: Any]?) async throws -> Any? {
-        let response = await AF.request("\(baseURLString)/\(url)", method: method, parameters: parameters, encoding: URLEncoding.default, headers: HTTPHeaders.default)
+        let encoding: ParameterEncoding = if method == .get { URLEncoding.default } else { JSONEncoding.default }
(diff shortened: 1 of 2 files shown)


theospears:
* `master`
  * **#437** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/437?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

PR #440, opened 2024-02-16
Update HealthKit on app launch

theospears:
Always update health metrics from apple before syncing with the server, for example on app launch. Because beeminder often update goals async, this doesn't guarantee that we will always show correct data reflecting the latest healthkit data, but should increase the chance. As the app polls for goals which are recalculating it means it should also show the correct data shortly after loading.

This might help a little with perceived Apple Health unreliabilty.

Testing:
Launched the app on a real device and observed data


theospears:
* `master`
  * **#440** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/440?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

PR #441, opened 2024-02-16
Ignore metadata when suggesting next value

theospears:
When choosing the default value to suggest for a new data point, the app looks at recent data points. Previously it would always choose the most recent value, but this was not useful if the most recent value was e.g. a derailment, or a pessimistic presumptive data point. Change the logic to now ignore these points when finding a value.

Testing:
Manually checked in the app with a goal with relevant data point
Wrote unit tests to extercise the various cases

### DIFF
--- a/BeeKit/Goal.swift
+++ b/BeeKit/Goal.swift
@@ -317,6 +317,12 @@ public class Goal {
         return self.autodata == "apple"
     }
 
+    /// A hint for the value the user is likely to enter, based on past data points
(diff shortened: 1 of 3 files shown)


theospears:
* `master`
  * **#441** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/441?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

PR #442, opened 2024-02-16
Goal view only sets suggested value once on load

theospears:
The goal screen tries to suggest a value based on recent past data.

In the past, this suggestion was updated every time the goal was updated/refreshed (as this might fetch new data). This lead to the user-selected value being overridden at annoying times, for example when navigating away from and returning to the app. Now instead we only set it once at load time. This might lead to stale values e.g. if goal data is out of date, but seems better than the previous behavior.

Testing:
Overrode the value, and checked pull to refresh did not change it
Checked switching away from the app and back did not change the value
Checked going back to the gallery and then back to the goal reset the value


theospears:
* `master`
  * **#442** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/442?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

<h2></h2>

PR #445, opened 2024-02-22
Dismiss the edit datapoint dialog after making a change

theospears:
Previously after updating or deleting a data point the dialog would just stay on the screen, until the user manually dismissed it. This was in part because otherwise after dismissing it would look like the user's change hadn't applied. The situation is better now because we show the lemniscate while there are pending goal updates.

Update the behavior to refresh goals after making a change, and then after the initial request (though likely with queuing) we will dismiss the dialog.

Testing:
Updated a data point and observed the goal refreshed and the dialog hid
Same for deleting a data point


theospears:
* **#445** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/445?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

PR #447, opened 2024-02-23
Authenticate against the beeminder api using HTTP Header

theospears:
The beeminder api now supports authenticating using the Authorization header, as an alternate to passing an access token as a parameter. Switch to using this form, as it keeps authentication information out of the URL, which means it is less likely to e.g. be leaked as part of logs, either on the client or server.

Testing:
* Verified read and write operations on the app worked
* Used debugger to check registering device token works
* Checked signing out and signing in still works


theospears:
* **#447** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/447?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

PR #448, opened 2024-02-29
Do not hard-crash on missing usernames

theospears:
Previously we were using `username!` to look up usernames. This would throw a fatal exception if the username was not available, crashing the app. In particular attempts to update HealthKit metrics could hit this after logging out from the app.

Avoid this by centralizing the username lookup logic in the request manager making URLs in the app slightly more readable, and have it throw a catchable exception. This could theoretically cause issues in the case that a URL contains an additional `{username}` string, but this seems unlikely.

Testing:
Launched the app on my phone and checked it could load goals.
Signed out and did not immediately experience a crash.

### DIFF
--- a/BeeKit/Managers/CurrentUserManager.swift
+++ b/BeeKit/Managers/CurrentUserManager.swift
@@ -110,7 +110,7 @@ public class CurrentUserManager {
         // Fetch a user from the persistent store
         let request = NSFetchRequest<User>(entityName: "User")
         // TODO: Handle (or at least log) an error here
-        let users = try? (context ?? container.viewContext).fetch(request)
+        let users = try? (context ?? container.newBackgroundContext()).fetch(request)
         return users?.first
     }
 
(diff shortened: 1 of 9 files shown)


theospears:
* **#448** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/448?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

BeeKit/Managers/CurrentUserManager.swift
```diff
@@ -110,7 +110,7 @@ public class CurrentUserManager {
         // Fetch a user from the persistent store
         let request = NSFetchRequest<User>(entityName: "User")
         // TODO: Handle (or at least log) an error here
-        let users = try? (context ?? container.viewContext).fetch(request)
+        let users = try? (context ?? container.newBackgroundContext()).fetch(request)
```
theospears, line 113: Drive by attempt to fix an unrelated crash

PR #468, opened 2024-07-28
Switch to use CoreData and remove old model

theospears:
This migrates usages across to the new CoreData model, and removes the old model. Keeping client code compatible with both was problematic, so instead do it in one go.

Testing:
Basic smoke testing. Likely has remaining bugs and will need further testing and fixes.


theospears:
* **#468** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/468?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -33,16 +33,15 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
     var lastUpdated: Date?
     let maxSearchBarHeight: Int = 50
     
-    var goals : Array<GoalProtocol> = []
-    var filteredGoals : Array<GoalProtocol> = []
+    var goals : Array<Goal> = []
+    var filteredGoals : Array<Goal> = []
```
krugerk, line 37: Since we are touching these lines anyway, how about adjusting them to this more modern syntax?

```suggestion
    var goals = [Goal]()
    var filteredGoals = [Goal]()
```

or

```suggestion
    var goals: [Goal] = []
    var filteredGoals: [Goal] = []
```

BeeKit/Model/DataPoint.swift
```diff
@@ -78,6 +78,7 @@ public class DataPoint: NSManagedObject {
         value = json["value"].numberValue
         comment = json["comment"].stringValue
         requestid = json["requestid"].stringValue
+        updatedAt = json["updated_at"].intValue
```
krugerk, line 81: This is meant to be called to update a datapoint's data. Shall we check that the ids match? That is expected, correct?

krugerk, line 81: How is updated_at different than last_touch?

I see updated_at for a goal at api.beeminder.com

> updated_at (number): Unix timestamp of the last time this goal was updated

and do not see last_touch mentioned there

krugerk: > lasttouch is meant to be more conservative. so it's the timestamp that the goal was touched in a "meaningful" way: data added or graph regenerated or... 
> updated_at is any time the goal was saved, including for reasons the user wouldn't care about

PR #475, opened 2024-09-14
CoreData objects track when last written to

theospears:
Adds lastModifiedLocal to all core data objects, which tracks when they were last modified. This is only a last local update, and does not e.g. correspond to when the server last changed.

This ensures that any write to these objects causes a change, making it easier to e.g. ensure we update Spotlight each time we write to the object, even if there are no property changes.


theospears:
* **#475** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/475?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

PR #479, opened 2024-09-21
Crash if failing to load core data

theospears:
Previously if there was an error loading core data (for example an incompatible schema) the app would wipe the db and retry. This was convenient for iteration, and might be good for robustness in production. But in beta/Testflight we want to catch and report errors, so we should just crash.

Testing
* [ ] Install the current release build, confirm we can upgrade to this build without crashing
* [ ] Install the current testflight build, confirm we can upgrade to this build without crashing


theospears:
* **#479** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/479?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `master`

<h2></h2>

PR #481, opened 2024-09-27
Bugfix 477 timer shows goal name

### LINKED ISSUE #477: Display goalname on timer screen
A User has requested that the timer screen be adjusted to include a reference to the corresponding goal. As is ([6.6](https://github.com/beeminder/BeeSwift/releases/tag/6.6)), the Timer View Controller covers the whole screen and includes only basics of the current timer: timer value, start/stop button, add datapoint, as well as exit and reset.

krugerk:
fixes #477 by displaying the name (_"slug"_) of the goal on the timer view controller

screenshot:

### DIFF
--- a/.github/workflows/fastlane-tests.yml
+++ b/.github/workflows/fastlane-tests.yml
@@ -9,7 +9,7 @@ concurrency:
 
 jobs:
   build:
-    runs-on: macos-14
+    runs-on: macos-15
 
     steps:
     - uses: maxim-lobanov/setup-xcode@v1
--- a/BeeSwift/GoalViewController.swift
+++ b/BeeSwift/GoalViewController.swift
@@ -322,9 +322,7 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
     }
 
     @objc func timerButtonPressed() {
-        let controller = TimerViewController()
-        controller.slug = self.goal.slug
-        controller.goal = self.goal
+        let controller = TimerViewController(goal: self.goal)
         controller.modalPresentationStyle = .fullScreen
         do {
             let hoursRegex = try NSRegularExpression(pattern: "(hr|hour)s?")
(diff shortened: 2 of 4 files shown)


theospears:
Thanks for making this fix. I also really appreciate you including a screenshot in the comment.

I have a few requests before merging:
1. This looks like it is 10% the feature change, and 90% a clean up. I'm on board with the cleanup, but could you put it in its own separate PR?
2. The change to button text really exacerbates the issue of buttons on this screen not looking like buttons. I've been trying to slowly move BeeSwift towards matching standard apple UI patterns. How would you feel about updating the button styles to be standard iOS buttons?
3. Please fix the test failures

krugerk:
> 1. This looks like it is 10% the feature change, and 90% a clean up. I'm on board with the cleanup, but could you put it in its own separate PR?

Thanks for the feedback! I can definitely split the cleanup into its own PR to keep things organized. I will get that sorted out and let you know once it is ready.

> 2. The change to button text really exacerbates the issue of buttons on this screen not looking like buttons. I've been trying to slowly move BeeSwift towards matching standard apple UI patterns. How would you feel about updating the button styles to be standard iOS buttons?

Thanks for pointing that out! I understand the importance of consistency with iOS UI patterns. I am on board with updating the button styles to align better with standard iOS buttons. I have made corresponding changes and have updated the branch and thus (currently) this PR.
At first I just adjusted them to look more like the one you had introduced in the edit data point screen:


And then used some inspiration from the bmndr website for the buttons - yellow on black instead of white on blue:

and





> 3. Please fix the test failures

None of the tests ran as the targeted device was not found. I bumped to macos15 and tweaked the commands in fastlane, which now build an iPhone 16 Pro Simulator with iOS 18.0 for the tests.

krugerk:
@theospears the Merge Request has been split and the buttons adjusted.

theospears (approved):

dreeves:
@krugerk those yellow-on-black buttons look very nice. thank you!

PR #483, opened 2024-09-30
Consistently styled buttons

krugerk:
Use a more consistent iOS styling for in app buttons.

Example:

See https://github.com/beeminder/BeeSwift/pull/481#issuecomment-2379771571

### DIFF
--- a/BeeKit/GoalExtensions.swift
+++ b/BeeKit/GoalExtensions.swift
@@ -59,15 +59,15 @@ extension Goal {
     public var countdownColor :UIColor {
         let buf = self.safeBuf
         if buf < 1 {
-            return UIColor.beeminder.red
+            return UIColor.Beeminder.red
         }
         else if buf < 2 {
-            return UIColor.beeminder.orange
+            return UIColor.Beeminder.orange
         }
         else if buf < 3 {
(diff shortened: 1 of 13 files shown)


theospears:
I think it's possible this style will be divisive - personally I find it quite loud. But I'm happy to put it in a TestFlight build and collect feedback. 

Could you update the PR to merge cleanly, now I have merged the test fixes in #481

theospears (approved):

theospears:
I've merged this because I don't think it's a great use of time to tease it apart, but to set expectations, in the future I will not merge PRs which combine behavior changes (like the styling here) and widespread code style changes (like UIColor.beeminder -> UIColor.Beeminder) in the same PR.

The reasons why are:
1. It makes it harder to review the PR - have to think carefully to decide whether changes are behavioral or not
2. It makes reverting harder if we decide to change the behavior, as it is more likely to conflict with later PRs

(I'm not looking for a discussion about whether or not this is the correct choice, just telling you the decisions and giving some context.)

theospears:
I think we should probably make this have a light background in light mode?

IMG_BE5BE99DAF80-1

PR #484, opened 2024-10-05
Target iOS 17

theospears:
This increases the minimum iOS version for BeeSwift to iOS17. Analytics data suggests that <2% of users are on iOS 16 or below. It also addresses the deprecation warnings associated with this new version.

Testing:
Confirmed it compiles


krugerk:
related: #623

PR #486, opened 2024-10-06
clean code

krugerk:
the verb to set up vs the noun the setup


theospears (approved):
The grammar is not important to me, but I agree being consistent with apple framework code (e.g. XCTestCase) makes sense.

PR #487, opened 2024-10-06
Clean code private set up

krugerk:
BSButton's setUp need not be exposed


krugerk:
(for after #486)

theospears:
I think @dreeves is working on graphite plan fixes so dealing with effectively stacked PRs like this is less painful

theospears (approved):
LGTM. Will merge after tests pass.

PR #489, opened 2024-10-06
Lighter button fill in light mode

krugerk:
Provides a lighter fill color for the button when in light mode and handles updating it dynamically as the appearance changes.

Examples:

### Light
Simulator Screenshot - Test-iPhone - 2024-10-06 at 14 09 30

### Dark
Simulator Screenshot - Test-iPhone - 2024-10-06 at 14 09 05

### DIFF
--- a/BeeKit/UI/BSButton.swift
+++ b/BeeKit/UI/BSButton.swift
@@ -13,11 +13,25 @@ public class BSButton : UIButton {
     
     required init?(coder aDecoder: NSCoder) {
         super.init(coder: aDecoder)
+        
+        registerForTraitChanges(
+            [UITraitUserInterfaceStyle.self]) {
+                (self: Self, previousTraitCollection: UITraitCollection) in
+                self.resetStyle()
+            }
+        
         self.setUp()
     }
     
     override init(frame: CGRect) {
         super.init(frame: frame)
+        
+        registerForTraitChanges(
+            [UITraitUserInterfaceStyle.self]) {
+                (self: Self, previousTraitCollection: UITraitCollection) in
+                self.resetStyle()
+            }
+        
         self.setUp()
     }
     
@@ -30,6 +44,10 @@ public class BSButton : UIButton {
         self.layer.borderColor = UIColor.Beeminder.yellow.cgColor
         self.layer.borderWidth = 1
         self.layer.cornerRadius = 4
+    }
+    
+    private func resetStyle() {
+        self.tintColor = dynamicTintFillColor
     }
     
     private let dynamicTintFillColor = UIColor { traitCollection in
--- a/BeeKit/UI/BSButton.swift
+++ b/BeeKit/UI/BSButton.swift
@@ -24,7 +24,7 @@ public class BSButton : UIButton {
     private func setUp() {
         self.titleLabel?.font = UIFont.beeminder.defaultBoldFont
         self.setTitleColor(UIColor.Beeminder.yellow, for: UIControl.State())
-        self.tintColor = .black
+        self.tintColor = dynamicTintFillColor
         self.configuration = .filled()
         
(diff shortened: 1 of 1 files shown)


krugerk:
contains changes from PRs #487 and #486

BeeKit/UI/BSButton.swift
```diff
@@ -13,23 +13,52 @@ public class BSButton : UIButton {
     
     required init?(coder aDecoder: NSCoder) {
         super.init(coder: aDecoder)
+        
+        registerForTraitChanges(
+            [UITraitUserInterfaceLevel.self]) {
```
theospears, line 18: Are you confident this is the correct trait to watch? https://developer.apple.com/documentation/uikit/uitraitcollection/3238085-userinterfacelevel doesn't make it sound particularly relevant to buttons.

BeeKit/UI/BSButton.swift
```diff
@@ -13,23 +13,52 @@ public class BSButton : UIButton {
...
     }
     
     override init(frame: CGRect) {
         super.init(frame: frame)
+        
+        registerForTraitChanges(
+            [UITraitUserInterfaceLevel.self]) {
+                (self: Self, previousTraitCollection: UITraitCollection) in
+                self.resetStyle()
+            }
```
theospears, line 33: As this is duplicated in both init methods, perhaps it should instead live in setUp?

BeeKit/UI/BSButton.swift
```diff
@@ -13,23 +13,52 @@ public class BSButton : UIButton {
...
+    private func resetStyle() {
+        self.tintColor = dynamicTintFillColor
+    }
+    
+    private let dynamicTintFillColor = UIColor { traitCollection in
+        switch traitCollection.userInterfaceStyle {
+        case .dark:
+            return UIColor.black
+        default:
+            return UIColor(red: 59.0/255.0,
```
theospears, line 58: I wonder what it would look like if we used the same grey as the +/- buttons. Possibly too low contrast with the yellow?

BeeKit/UI/BSButton.swift
krugerk: UITraitUserInterfaceStyle looks to be more on point.

BeeKit/UI/BSButton.swift
krugerk: I think you are right. The contrast seems to work for me on the sign in button but looks too low contrast elsewhere.

How about something like this - yellow on black in dark and black on grey in light mode?

https://github.com/user-attachments/assets/34a7d634-3826-45d5-ab82-a02c95e37981

theospears: What if we skipped the border entirely and just had the solid button, either gray or yellow?

theospears: Grey (OK):
Screenshot 2024-10-06 at 2 49 29 PM


Yellow + Black Text (Probably too loud):
Screenshot 2024-10-06 at 2 50 42 PM

Yellow + White Text (Unreadable):
Screenshot 2024-10-06 at 2 51 14 PM

theospears: And for completeness, default system blue:
Screenshot 2024-10-06 at 2 53 23 PM

krugerk: And the current target? Yellow text on Black background with yellow border in dark mode and black text on grey and without border in light mode?

theospears: My take would be:
* Yellow border and text, black background when shown in dark mode
* Black on grey without border in light mode

But I don't feel like I have particularly good graphics design taste, so I'm happy for you to make a call on what you want to try, and we can put it out to test flight to get wider feedback.

PR #490, opened 2024-10-08
HKCategoryValueSleepAnalysis.asleep deprecated on ios16

krugerk:
This is a follow-up to Commit e5ef0f / MR #484.

It replaces uses of `asleep` with `asleepUnspecified` in the other targets (tests).
```diff
-  HKCategoryValueSleepAnalysis.asleep
+  HKCategoryValueSleepAnalysis.asleepUnspecified
```


theospears:
Thanks for fixing this!

PR #492, opened 2024-10-14
Black on grey without border in light mode

krugerk:
follow-up to #489, implementing https://github.com/beeminder/BeeSwift/pull/489#discussion_r1792894445

### light
light - sign in

### dark
dark - sign in


krugerk:
As a note to self, now that the app is iOS 15 and up, we could look into utilizing [ UIButton.Configuration](https://developer.apple.com/documentation/uikit/uibutton/configuration).

theospears (approved):

PR #494, opened 2024-10-15
gave notifications vc progressHUD, PTR, and section titles

krugerk:
user can now manually trigger a refresh of the notification settings using the standard 'pull to refresh' mechanism

also, the goals are sorting a to z rather than z to a and they are grouped by those using the defaults and those not

### DIFF
--- a/BeeSwift/Settings/ConfigureNotificationsViewController.swift
+++ b/BeeSwift/Settings/ConfigureNotificationsViewController.swift
@@ -131,19 +131,25 @@ extension ConfigureNotificationsViewController : UITableViewDataSource, UITableV
     }
     
     func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
-        if section == 0 { return 1 }
-        if section == 1 { return self.goalsUsingDefaultNotifications.count }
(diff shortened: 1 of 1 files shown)


krugerk:
@theospears if you would like I could split this into either separate commits (where one can step through commits to see the atomic changes) or separate merge requests

BeeSwift/Settings/ConfigureNotificationsViewController.swift
```diff
@@ -78,47 +84,113 @@ class ConfigureNotificationsViewController: UIViewController {
...
     func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
-        if section == 1 { return self.goals.count }
-        return 1
+        switch section {
+        case 0:
+            return 1
+        case 1:
+            return self.goalsUsingDefaultNotifications.count
+        default:
+            return self.goalsUsingNonDefaultNotifications.count
```
theospears, line 140: Non blocking, but I'm a fan of explicitly handling all the expected cases and have `default:` log/report an error.

theospears (approved):

Issue #64, opened 2020-06-15
Different Layout for the iPad app


krugerk:
This is a feature request found in the [Beeminder Forums](https://forum.beeminder.com/t/different-layout-for-the-ipad-app/6400).

PR #495, opened 2024-10-20
goalVC now also respects safe area

krugerk:
goal view controller was not respecting the safe areas. This meant, for example, that the graph was partially hidden behind the Dynamic Island.

### DIFF
--- a/BeeSwift/GoalViewController.swift
+++ b/BeeSwift/GoalViewController.swift
@@ -62,12 +62,11 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
 
         self.view.addSubview(self.scrollView)
         self.scrollView.snp.makeConstraints { (make) -> Void in
-            make.top.equalToSuperview()
-            make.left.equalTo(0)
-            make.right.equalTo(0)
-            make.bottom.equalTo(0)
+            make.top.equalTo(self.view.safeAreaLayoutGuide.snp.topMargin)
+            make.left.equalTo(self.view.safeAreaLayoutGuide.snp.leftMargin)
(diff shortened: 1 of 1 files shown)


theospears:
I've really appreciated the before/after screenshots you've included in some other PRs - could you add one here?

krugerk:
> I've really appreciated the before/after screenshots you've included in some other PRs - could you add one here?

sure.

This is what it looks like broken / currently:

safe area left
safe area right

The data area though on the goal vc accounted for the [safe area](https://developer.apple.com/documentation/uikit/uiview/positioning_content_relative_to_the_safe_area):
data area already respecting left and right safe area

thanks to https://github.com/beeminder/BeeSwift/blob/a4b2744e67f7dbfe1e6d15eb3adc6a8a12327b55/BeeSwift/GoalViewController.swift#L132-L133 and https://github.com/beeminder/BeeSwift/blob/a4b2744e67f7dbfe1e6d15eb3adc6a8a12327b55/BeeSwift/GoalViewController.swift#L105-L106

theospears:
This seems reasonable, although I wonder if the app should even support landscape mode on phones

krugerk:
> This seems reasonable, although I wonder if the app should even support landscape mode on phones

Yeah that could be worth the discussion. This has been broken since iPhone X! (Some apps do purposely have content scroll inside the safe area and behind the Dynamic Island or into the status bar area.)
The app is (still) somewhat inconsistent about these sorts of things. Currently I am thinking we just iteratively clean it up, at least until the sweeping overhaul.

krugerk:
mentioning #64, a table exclusive layout

PR #496, opened 2024-10-21
date picker wraps around to 1 for the first of the current month

### LINKED ISSUE #386: Date picker wraps around to 1 for the first of the current month
```[tasklist]
### Desiderata
- [ ] Add the special case or whatever?
- [ ] Discuss how to totally redesign this to make it moot?
```

## Replicata 

1. Say it's May 31st and you want to enter data for tomorrow
2. So you hit the "+" on the "31" and it wraps around to "1"
3. So far so good, seemingly
4. Submit the datapoint

## Expectata

A datapoint added for June 1st.

## Resultata

A datapoint added for May 1st!

If this is a do-less goal, this can even derail you.

(Pretty bad for a typical cumulative do-more goal as well since your data will just be all wrong but you won't notice since it will just bump up your current value like normal even though you technical added the datapoint a month in the past.)

_**Verbata**: anti-magic, date wrapping, datepicker, date-picking, specifying datapoint date as day-of-month,_

krugerk:
The Edit Datapoint screen contains a date picker. The Add Datapoint screen, the lower part of the goal screen, did not contain a date picker. Instead it contained a single digit to represent the date of the to-be-created datapoint and furthermore sent this single digit to the backend as part of `urtext`. The backend itself then interpreted the values sent and really had no way to accurately map the "day of the month" number to any particular year or month.

With this merge request, the Add Datapoint section has been given its own date which is then sent to the backend when adding the new datapoint. Since the app now provides all of year, month, and day, there is less ambiguity related to the daystamp for the datapoint.

Furthermore, similar to how the [Android app / Beedroid](https://github.com/beeminder/BeeSwift/issues/386#issuecomment-1574351335) handles this, the format shown varies: the (date) value shown is either a simple single number (the day) when within the same month, the month and day when no longer in the same month but still the same year, and all three components when the point is for a different calendar year.

example, stepping through dates on 2024-Oct-21:

https://github.com/user-attachments/assets/fb4f0d23-6880-4e8d-97a9-4ed38dff4185

Fixes #386

### DIFF
--- a/BeeSwift/GoalViewController.swift
+++ b/BeeSwift/GoalViewController.swift
@@ -457,14 +457,7 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
     }
     
     private var urtext: String {
-        let urtextDateFormatter: DateFormatter = {
-            let formatter = DateFormatter()
-            formatter.locale = Locale(identifier: "en_US")
-            formatter.dateFormat = "yyyy MM dd"
-            return formatter
-        }()
-        
-        return "\(urtextDateFormatter.string(from: self.date)) \(self.valueTextField.text!) \"\(self.commentTextField.text!)\""
+        return "\(DateFormatter.urtextDateString(from: self.date)) \(self.valueTextField.text!) \"\(self.commentTextField.text!)\""
     }
 
     @objc func submitDatapoint() {
@@ -530,5 +523,19 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
 
     func safariViewControllerDidFinish(_ controller: SFSafariViewController) {
         controller.dismiss(animated: true, completion: nil)
+    }
+}
+
+
+private extension DateFormatter {
+    private static let urtextDateFormatter: DateFormatter = {
+            let formatter = DateFormatter()
+            formatter.locale = Locale(identifier: "en_US")
+            formatter.dateFormat = "yyyy MM dd"
+            return formatter
+        }()
+    
+    static func urtextDateString(from date: Date) -> String {
+        urtextDateFormatter.string(from: date)
     }
 }
--- a/BeeSwift/GoalViewController.swift
+++ b/BeeSwift/GoalViewController.swift
@@ -275,7 +275,7 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
         }
 
         self.navigationItem.rightBarButtonItems = [UIBarButtonItem(barButtonSystemItem: .action, target: self, action: #selector(self.actionButtonPressed))]
-        if (!self.goal.hideDataEntry) {
+        if !self.goal.hideDataEntry {
             self.navigationItem.rightBarButtonItems?.append(UIBarButtonItem(image: UIImage(named: "Timer"), style: .plain, target: self, action: #selector(self.timerButtonPressed)))
         }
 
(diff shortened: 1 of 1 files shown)


BeeSwift/GoalViewController.swift
```diff
@@ -295,7 +298,7 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
         }
 
         self.navigationItem.rightBarButtonItems = [UIBarButtonItem(barButtonSystemItem: .action, target: self, action: #selector(self.actionButtonPressed))]
-        if (!self.goal.hideDataEntry) {
+        if !self.goal.hideDataEntry {
             self.navigationItem.rightBarButtonItems?.append(UIBarButtonItem(image: UIImage.init(named: "Timer"), style: .plain, target: self, action: #selector(self.timerButtonPressed)))
```
krugerk, line 302:
```suggestion
            self.navigationItem.rightBarButtonItems?.append(UIBarButtonItem(image: UIImage(named: "Timer"), style: .plain, target: self, action: #selector(self.timerButtonPressed)))
```

BeeSwift/GoalViewController.swift
```diff
@@ -471,9 +476,9 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
         }
         return true
     }
-
-    func urtextFromTextFields() -> String {
-        return "\(self.dateTextField.text!) \(self.valueTextField.text!) \"\(self.commentTextField.text!)\""
+    
+    private var urtext: String {
+        return "\(DateFormatter.urtextDateString(from: self.date)) \(self.valueTextField.text!) \"\(self.commentTextField.text!)\""
```
krugerk, line 481:
```suggestion
       "\(DateFormatter.urtextDateString(from: self.date)) \(self.valueTextField.text!) \"\(self.commentTextField.text!)\""
```

BeeSwift/GoalViewController.swift
krugerk: see merge request: #501

BeeSwift/GoalViewController.swift
```diff
@@ -368,12 +371,14 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
         var components = DateComponents()
         components.day = Int(self.dateStepper.value)
 
-        let newDate = (calendar as NSCalendar?)?.date(byAdding: components, to: Date(), options: [])
+        let now = Date()
+        guard let newDate = calendar.date(byAdding: components, to: now) else { return }
```
theospears, line 375: Nit: This seems like the return here should never happen. I like logging a warning in these cases.

BeeSwift/GoalViewController.swift
```diff
@@ -368,12 +371,14 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
         var components = DateComponents()
         components.day = Int(self.dateStepper.value)
 
-        let newDate = (calendar as NSCalendar?)?.date(byAdding: components, to: Date(), options: [])
+        let now = Date()
```
theospears, line 374: This does surprising things if you have the view open and midnight rolls around. But this PR doesn't change that weird behavior, it was always there.

theospears (approved):
Looks good to me. Thanks for including the video.

BeeSwift/GoalViewController.swift
krugerk: True. The app will work with the data it has from the time it last fetched it. One can use other means to adjust the deadline and the one client might not know about it while it is at the timer screen, add data screen, or edit data screen.

Is the server posting to the app via remote notification that a goal has been updated? Does the app still need to fetch (pull) to find out?

PR #497, opened 2024-10-21
timer screen figures out its own units

krugerk:
The Timer Screen has special logic for interpreting the accumulated seconds, dividing the value by 60 when the goal's unit is (perhaps, probably) in minutes

Previously the goal screen included the logic for figuring out which units the timer screen should use.

### DIFF
--- a/BeeSwift/TimerViewController.swift
+++ b/BeeSwift/TimerViewController.swift
@@ -14,7 +14,7 @@ import BeeKit
 
 class TimerViewController: UIViewController {
     private enum TimerUnit {
-        case hours, minutes, other
+        case hours, minutes
     }
 
     let timerLabel = BSLabel()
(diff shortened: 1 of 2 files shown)


BeeSwift/TimerViewController.swift
```diff
@@ -199,3 +211,19 @@ class TimerViewController: UIViewController {
         }
     }
 }
+
+private extension TimerViewController {
+    static private func timerUnit(goal: Goal) -> TimerUnit {
```
theospears, line 216: Random musing:
Another option for representation could be
1) Make TimerUnit only have actual time units:
```
private enum TimerUnit {
        case hours, minutes
}
```

2) Have this function return an optional result 
```
static private func timerUnit(goal: Goal) -> TimerUnit?
```

3) Have the call site fall back to hours:

```
self.units = Self.timerUnit(goal: goal) ?? .hours
```

This might reduce the number of places in the code which need to understand to treat `other` and `hours` the same.

Issue #114, opened 2020-07-31
support configuring the today widget


krugerk:
[tentatively nixed per the anti-settings principle but if people have use cases they can describe, they should!]

The shortcuts app allows one to exclude a shortcut from the widget.
There is a screen to move them around as well as toggle inclusion.
And one can toggle the widget inclusion at each shortcut.

The beeminder app does not support configuring it but I was thinking "always include" and "never include" might be valid options when configuring the today widget.

Issue #98, opened 2020-07-24
Today Widget goals not updated after entering data via Today Widget


krugerk:
```[tasklist]
### Desiderata
- [ ] Confirm replicata for version 6.5 of the app
- [ ] Discuss whether this is actually bad (not obvious to dreev that it is!)
- [ ] TBD
```

Currently the widget displays at most one goal.

## Replicata

1. Enter data via the widget such that that goal is no longer the most urgent
2. Goal is updated, this works.

## Expectata

That the new most urgent goal is shown in the widget

## Resultata

The goal we just added data to is still the one listed.

## Workaround

Open the app, fetch the goals, return to the Today Widget.

_**Verbata**: staleness, freshness, syncing with the server,_

krugerk:
5.7 (2)

The graph for the goal is updated but the description of the amount due is not.

'Add data' in today cell requests a new graph, which is displayed once it is fetched. This part works correctly.

In fact, if the widget is removed from the today screen then added again - without opening the beeminder app directly, the today widget shows old data.

krugerk:
Right now (5.7) the widget relies on the main app for its data. The today widget can submit new data but that new data will not be reflected in the today widget until the main app has become aware of the new data and subsequently updated the data used for the today widget.

theospears:
Today widgets are deprecated, and will see no further development

Issue #97, opened 2020-07-24
Entering app from widget does not always take you to that goal's screen


krugerk:
```[tasklist]
### Desiderata
- [ ] Is this still relevant? Is the Today widget still a thing?
- [ ] Do the replicata still replicate?
- [ ] TBD
```

## Replicata from krugerk

1. Be on version 5.6.4 or later?
2. Log in
3. Configure the Beeminder Today widget (add the widget to the Today screen)
4. Force-close the app
5. View the Today widget, tap a goal

## Expectata

To see the goal screen for the goal you tapped.

## Resultata

App opens, displays gallery.

## Nota Nebulosa

Tapping/selecting a row in the Today widget results in [a beeminder:// URL being created](https://github.com/beeminder/BeeSwift/blob/fcc79a30b8655ceb9ce66dde21d7c7cd3eef08b5/BeeSwiftToday/TodayViewController.swift#L66). The App is registered to process this and this much works -- the app starts. 

It could be a timing issue in that the app sends the "open goal" request before the goals have been loaded from the backend/server.

## Cognata

* #41 

_**Verbata**: today widget, user's intended destination,_

krugerk:
Only partially related to #41 in that there is currently no code for #41 attempting to open the goal from a remote notification.

theospears:
Today widgets are deprecated, and won't get any additions or fixes. As/when new style widgets are built we can work on this.

PR #500, opened 2024-10-23
removes the deprecated and unmaintained Today View extension

krugerk:
Today View extensions have been deprecated since iOS 14. Furthermore they are [no longer available at all on iOS 18](https://developer.apple.com/documentation/ios-ipados-release-notes/ios-ipados-18-release-notes#Widgets).

This app's such extension allowed for viewing up to three goals and provided quick data entry for them.
It seems each iteration of iOS since 14 has tucked the Today View extensions further and further out of reach.
A more powerful replacement could be iOS 14 style widgets (#68).

Today View extension tickets: #97, #98, #114, #236

This merge request merely removes the legacy code/feature.


theospears:
I was thinking of leaving this in until new widgets are ready, but don't feel strongly about it to open to being persuaded it should just be removed now.

theospears:
Given this is gone in iOS 18, let's rip it out

theospears (approved):

PR #501, opened 2024-10-24
clean code - removed explicit init where not required

krugerk:
removed explicit init where not required


krugerk:
This is coding style/convention. [SwiftLint](https://github.com/realm/SwiftLint) and/or [SwiftFormat](https://github.com/nicklockwood/SwiftFormat) can take care of this for us.

theospears (approved):

PR #503, opened 2024-10-24
avoid keyboard overlapping gallery cells

### LINKED ISSUE #200: While searching last goal is covered by keyboard

krugerk:
adjusts the gallery's insets to account for the keyboard

fixes #200

(and replaces #275)

The app does not yet support tags (filtering/sorting/setting) and it might be common practice to instead use the search/filter to have goals with particular characters in their names appear in the gallery. When doing so, the soft keyboard appears and potentially obscures the bottom most two or so goal cells.

### DIFF
--- a/BeeSwift/Gallery/GalleryViewController.swift
+++ b/BeeSwift/Gallery/GalleryViewController.swift
@@ -43,6 +43,8 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
         NotificationCenter.default.addObserver(self, selector: #selector(self.handleSignOut), name: NSNotification.Name(rawValue: CurrentUserManager.signedOutNotificationName), object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(self.openGoalFromNotification(_:)), name: NSNotification.Name(rawValue: "openGoal"), object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(self.handleGoalsFetchedNotification), name: NSNotification.Name(rawValue: GoalManager.goalsUpdatedNotificationName), object: nil)
+        NotificationCenter.default.addObserver(self, selector: #selector(keyboardWillHide), name: UIResponder.keyboardWillHideNotification, object: nil)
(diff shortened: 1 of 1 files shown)


theospears:
`IQKeyboardManager` is meant to magically deal with this, and I'd rather use it over writing our own custom logic if possible. Any idea why it isn't working in this case?

krugerk:
> `IQKeyboardManager` is meant to magically deal with this, and I'd rather use it over writing our own custom logic if possible. Any idea why it isn't working in this case?

As far as I can tell IQKeyboardManager is meant to magically deal with preventing the onscreen keyboard from covering `UITextView` or `UITextField`. Maybe IQKeyboardManager does not handle `UISearchBar`, but then again the search bar is not being covered by the keyboard in this case.

krugerk:
@theospears what is missing in this one?

theospears:
How do you feel about the approach in #536 ?

krugerk:
#536 for the win!

PR #506, opened 2024-10-28
hint for no goal when filtering and ptr on empty list

### LINKED ISSUE #226: gallery without goals cannot be refreshed directly
For accounts without active goals, the gallery (main app interface) shows a message:

> No goals yet!
>
> In-app goal creation is coming soon, but for now,
> head to bee minder.com to create a goal.

This message appears for new accounts and older accounts, in either case when there are no active goals.

So the user opens a browser and opens the beeminder website. At the website, the user creates a goal. Finished. The user returns to the iOS app and still sees the gallery above. Given the app is an iOS app, the user attempts a pull-to-refresh, which appears to have no effect. 

Browsing through the code we know the app will fetch goals again once the gallery is shown again but for the user at this screen, it still shows 'no goals yet' although the user just created one.

If the app is not going to receive a push message that a new goal has been created, the gallery when empty should still perhaps allow pull-to-refresh or another means for allowing the user to directly request a reload.

krugerk:
uses the common technique of setting the backgroundView of the CollectionView when no items would be displayed

fixes #226

new notice screen:
Simulator Screenshot - iPhone Xs - no beeminder goals notice

and now with pull-to-refresh:
Simulator Screenshot - iPhone Xs - no beeminder goals notice - pull to refresh

and the new hint:
Simulator Screenshot - iPhone Xs - beeminder goals but none showing due to filter


BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -431,9 +414,19 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
     }
     
     func numberOfSections(in collectionView: UICollectionView) -> Int {
-        return 1
+        switch (filteredGoals.isEmpty, goals.isEmpty) {
```
theospears, line 417: `numberOfSections` seems like it should be a pure function which just returns a number and doesn't have side effects like updating the background. Maybe `updateGoals()` is a better place for this to live?

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -519,3 +512,57 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
     }
 }
 
+private extension GalleryViewController {
```
theospears, line 515: This seems like a good candidate to be its own view, rather than an extension to `GalleryViewController`

BeeSwift/Gallery/GalleryViewController.swift
krugerk: I am not sure I understand. With this change the background is set according to the number of sections and is being set in the numberOfSections. So yeah, that method has the side effect of setting the background. 
As is, `updateGoals()` does not seem to be an ideal spot for the change either.
I made a change; take a look

BeeSwift/Gallery/GalleryViewController.swift
krugerk: What did you have in mind? It is just a label with a bit of spacing, configured for being displayed in the Gallery. It is here inside this `private extension` so we can keep the details out of the higher level further up where it is used.
To better follow the file's pattern, it could be configured in the viewDidLoad.
I think if it were in SwiftUI, it would be its own named view.

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -397,25 +395,31 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
...
-            self.noGoalsLabel.isHidden = true
-            self.collectionView?.isHidden = false
-        }
-        let searchItem = UIBarButtonItem(barButtonSystemItem: .search, target: self, action: #selector(self.searchButtonPressed))
-        self.navigationItem.leftBarButtonItem = searchItem
+        self.navigationItem.leftBarButtonItem = UIBarButtonItem(barButtonSystemItem: .search, target: self, action: #selector(self.searchButtonPressed))
     }
 
     func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
         return ServiceLocator.versionManager.lastChckedUpdateState() == .UpdateRequired ? 0 : self.filteredGoals.count
```
krugerk, line 402: - [ ] include 'update required' as one of the reasons for an empty list (show need upgrade in the background view)

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -408,25 +406,31 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
...
     }
     
     func numberOfSections(in collectionView: UICollectionView) -> Int {
-        return 1
+        [filteredGoals.isEmpty, goals.isEmpty].contains(true) ? 0 : 1
```
theospears, line 417:
```suggestion
        filteredGoals.isEmpty ? 0 : 1
```
If `goals` is empty then filtered goals will always be, right?

BeeSwift/Gallery/GalleryViewController.swift
theospears: I'm pretty -1 on extensions overall - I see them as resulting in code being in multiple places, without adding anything in the way of real structure. Having said that, I'm not going to totally veto them, but as the very least each one needs a comment describing the functionality it contains.

theospears: [Do] Add a comment summarizing the functionality encased in this extension.

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -27,14 +27,21 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
...
-    let noGoalsLabel = BSLabel()
     let outofdateLabel = BSLabel()
     let searchBar = UISearchBar()
     var lastUpdated: Date?
     let maxSearchBarHeight: Int = 50
     
-    var goals : [Goal] = []
-    var filteredGoals : [Goal] = []
+    var goals : [Goal] = [] {
+        didSet {
```
theospears, line 36: [Do] Move this logic (and filteredGoals.didSet) to `didUpdateGoals` which contains all the other logic/view invalidation which is run when the set of goals update.

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -408,25 +406,31 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
...
     func numberOfSections(in collectionView: UICollectionView) -> Int {
-        return 1
+        [filteredGoals.isEmpty, goals.isEmpty].contains(true) ? 0 : 1
     }
     
+    private func setBackgroundView(filteredGoals: [Goal], goals: [Goal]) {
+        var view: UIView? {
+            switch (filteredGoals.isEmpty, goals.isEmpty) {
+            case (true, false):
+                return self.makeViewForEmptyCollection(when: .noGoalsMatchingFilter)
```
theospears, line 424: [Do] Remove the enum and the multiple views, just pass in a string here.

I've marked this as a Do, but I'm open to being convinced otherwise. At the moment I don't see the benefit to justify the additional complexity.

theospears (changes requested):
I agree using BackgroundView on CollectionView is a reasonable way to handle this. 

This PR will conflict with #542. Given you opened it first I don't want to put the work on fixing all the merge conflicts on you, so added some feedback about what updates are necessary for this to be merged. If you're happy to follow up on them in the next week I'll hold off on #542 and fix it up to be compatible after this is merged.

krugerk:
> I agree using BackgroundView on CollectionView is a reasonable way to handle this.
> 
> This PR will conflict with #542. Given you opened it first I don't want to put the work on fixing all the merge conflicts on you, so added some feedback about what updates are necessary for this to be merged. If you're happy to follow up on them in the next week I'll hold off on #542 and fix it up to be compatible after this is merged.

Thanks for the consideration. (some amount of) Updating the MR so that it is mergable with its target is to be expected. And I would update it accordingly.

BeeSwift/Gallery/GalleryViewController.swift
krugerk: Yes, if `goals` is empty, then any subset of it would also be empty, including the one known as `filteredGoals`.

I see what you mean: we are always using the combination of the filter applied to the goals and thus the set in `filteredGoals`. When the searchBar.text the filtered is considered off and all goals are in `filteredGoals`.

theospears: Perhaps these names are unclear. Something like `allGoals` and `visibleGoals` might have been a better choice?

BeeSwift/Gallery/GalleryViewController.swift
krugerk: Currently `collectionView?.backgroundView` is being set to a UIView which is just a container, containing a BSLabel with the reason no goals are shown, and, below it, another UIView acting as a spacer.

The intent was to show the text/label "above the fold" for better visibility and to not have it sit in the middle, which would be covered by the MBProgressHUD activity indicator.

Your proposal is just to set the backgroundView to the label directly?

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -499,3 +500,57 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
     }
 }
 
+private extension GalleryViewController {
```
krugerk, line 503:
```suggestion
/// factory to create the view for use in the collectionView's background when no goals are shown
private extension GalleryViewController {
```

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -441,20 +442,20 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
...
         return cell
     }
     
     func collectionView(_ collectionView: UICollectionView, layout collectionViewLayout: UICollectionViewLayout, referenceSizeForFooterInSection section: Int) -> CGSize {
-        return CGSize(width: 320, height: section == 0 && self.goals.count > 0 ? 5 : 0)
+        return CGSize(width: 320, height: section == 0 && self.allGoals.count > 0 ? 5 : 0)
     }
     
     func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
         let row = (indexPath as NSIndexPath).row
```
krugerk, line 457:
```suggestion
        let row = indexPath.row
```

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -441,20 +442,20 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
...
-        if row < self.filteredGoals.count { self.openGoal(self.filteredGoals[row]) }
+        if row < self.visibleGoals.count { self.openGoal(self.visibleGoals[row]) }
```
krugerk, line 458:
```suggestion
        if row < self.visibleGoals.count {
            self.openGoal(self.visibleGoals[row])
        }
```

BeeSwift/Gallery/GalleryViewController.swift
theospears: I'm happy either to set it to the view, or the label directly. I just don't think we get much from having the `NoGoalReason` enum - `makeViewForEmptyCollection` could just take the string to show.

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -463,7 +450,7 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
             }
         }
         else if let slug = notif.userInfo?["slug"] as? String {
-            matchingGoal = self.goals.filter({ (goal) -> Bool in
+            matchingGoal = self.allGoals.filter({ (goal) -> Bool in
                 return goal.slug == slug
             }).last
```
krugerk, line 455:
```suggestion
            matchingGoal = self.allGoals.last(where: { $0.slug == slug })
```

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -485,3 +472,60 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
...
+// view for the background of the collection view based on its contents
+private extension GalleryViewController {
+    var backgroundViewConsideringContents: UIView? {
+        switch (visibleGoals.isEmpty, allGoals.isEmpty) {
+        case (true, false):
+            return self.makeViewForEmptyCollection(when: .noGoalsMatchingFilter)
+        case (true, true):
+            return self.makeViewForEmptyCollection(when: .noActiveGoals)
+        default:
+            return nil
```
krugerk, line 485:
```suggestion
        case (true, false):
            self.makeViewForEmptyCollection(when: .noGoalsMatchingFilter)
        case (true, true):
            self.makeViewForEmptyCollection(when: .noActiveGoals)
        default:
            nil
```

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -429,20 +416,20 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
     func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
         let cell:GoalCollectionViewCell = self.collectionView!.dequeueReusableCell(withReuseIdentifier: self.cellReuseIdentifier, for: indexPath) as! GoalCollectionViewCell
         
-        let goal:Goal = self.filteredGoals[(indexPath as NSIndexPath).row]
+        let goal:Goal = self.visibleGoals[(indexPath as NSIndexPath).row]
```
krugerk, line 419:
```suggestion
        let goal:Goal = self.visibleGoals[indexPath.row]
```

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -429,20 +416,20 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
     func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
         let cell:GoalCollectionViewCell = self.collectionView!.dequeueReusableCell(withReuseIdentifier: self.cellReuseIdentifier, for: indexPath) as! GoalCollectionViewCell
```
krugerk, line 417: why `self.collectionView` with `!` and not merely the `collectionView` passed into this method?

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -380,28 +370,25 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
...
     }
 
     func collectionView(_ collectionView: UICollectionView, numberOfItemsInSection section: Int) -> Int {
-        return versionManager.lastChckedUpdateState() == .UpdateRequired ? 0 : self.filteredGoals.count
+        return versionManager.lastChckedUpdateState() == .UpdateRequired ? 0 : self.visibleGoals.count
     }
     
     func numberOfSections(in collectionView: UICollectionView) -> Int {
-        return 1
+        [visibleGoals.isEmpty, allGoals.isEmpty].contains(true) ? 0 : 1
```
krugerk, line 389:
```suggestion
        visibleGoals.isEmpty ? 0 : 1
```

BeeSwift/Gallery/GalleryViewController.swift
krugerk: unrelated to this PR though
```swift
 @objc func openGoalFromNotification(_ notification: Notification) {
        var goalFromIdentifier: Goal? {
            guard
                let identifier = notification.userInfo?["identifier"] as? String,
                let url = URL(string: identifier),
                let objectID = viewContext.persistentStoreCoordinator?.managedObjectID(forURIRepresentation: url)
            else { return nil }
                return viewContext.object(with: objectID) as? Goal
            }
        
        var goalFromSlug: Goal? {
            guard
                let slug = notification.userInfo?["slug"] as? String
            else { return nil }
            return self.allGoals.last(where: { $0.slug == slug })
        }
        
        if let matchingGoal = goalFromIdentifier ?? goalFromSlug {
            self.navigationController?.popToRootViewController(animated: false)
            self.openGoal(matchingGoal)
        }
    }
```

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -36,14 +36,13 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
...
-    let noGoalsLabel = BSLabel()
     let outofdateLabel = BSLabel()
     let searchBar = UISearchBar()
     var lastUpdated: Date?
     let maxSearchBarHeight: Int = 50
     
-    var goals : [Goal] = []
-    var filteredGoals : [Goal] = []
+    var allGoals : [Goal] = []
+    var visibleGoals : [Goal] = []
```
krugerk, line 45:
```suggestion
    var allGoals = [Goal]()
    var visibleGoals = [Goal]()
```

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -36,14 +36,13 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
     let cellReuseIdentifier = "Cell"
     var deadbeatView = UIView()
     var outofdateView = UIView()
```
krugerk, line 38:
```suggestion
    let deadbeatView = UIView()
    let outOfDateView = UIView()
```

Issue #268, opened 2021-02-23
the in-app timer keep the screen on/awake


krugerk:
A user in the [forum](https://forum.beeminder.com/t/feature-request-in-app-timer-keeps-screen-on-awake/9033) has requested that the in-app timer keep the screen on/awake.

> I’d like to request that the in-app timer have the option to keep the screen awake, either by default if the screen is on Beeminder with the in-app timer running (preferred!) or as a user-configured setting (but why complicate matters? you can always override by navigating away from the app or manually turning off your screen).
> 
> Looking up and seeing the timer counting is a great way to stay on track AND make sure you stop the timer when you finish.
> 
> I’m on iOS, if it makes a difference.

I would add to this, a local notification may also be useful. It would indicate the goal, elapsed time, and  start time. Obviously tapping the notification should take the user to the running timer screen in the app.

PR #508, opened 2024-11-06
keeping the timer screen on with timer running

krugerk:
keep screen on, albeit dimmed, with active timer on timer screen

fixes: #268


BeeSwift/TimerViewController.swift
```diff
@@ -9,18 +9,26 @@
...
     let goal: Goal
     var timingSince: Date?
     var timer: Timer?
     var units: String?
     var accumulatedSeconds = 0
-
+    
+    private var idleTimer: Timer?
+    private var defaultBrightnessLevel: CGFloat = UIScreen.main.brightness
+    private let dimmedBrightnessLevel: CGFloat = 0.2
```
theospears, line 29: Perhaps this should be a function of the default brightness level? If the user has brightness turned low this could end up increasing it, right?

BeeSwift/TimerViewController.swift
```diff
@@ -9,18 +9,26 @@
...
     let startStopButton = BSButton(type: .system)
     let goal: Goal
     var timingSince: Date?
     var timer: Timer?
     var units: String?
     var accumulatedSeconds = 0
-
+    
+    private var idleTimer: Timer?
+    private var defaultBrightnessLevel: CGFloat = UIScreen.main.brightness
```
theospears, line 28: We should probably capture the current value as late as possible (i.e. just before changing it), to minimize surprising behavior if the user changes brightness while on this screen?

BeeSwift/TimerViewController.swift
```diff
@@ -90,6 +98,16 @@ class TimerViewController: UIViewController {
         resetButton.setTitle("Reset", for: .normal)
     }
     
+    override func viewWillDisappear(_ animated: Bool) {
```
theospears, line 101: In addition to when leaving the screen, should we disable the brightness override when the app is moved to background?

BeeSwift/TimerViewController.swift
```diff
@@ -125,6 +147,8 @@ class TimerViewController: UIViewController {
             self.timer = nil
             self.timingSince = nil
         }
+        
+        resetIdleTimer()
```
theospears, line 151: The lifecycle of "when we disable the screen" is a little surprising to me. In particular, we enter "do not sleep" mode when you start the timer, but do not leave it when you stop the timer. I'd suggest either enabling it as soon as the user enters the screen, or disabling it when the timer is not running.

PR #509, opened 2024-11-06
upgrade gems

krugerk:
mostly to bump from REXML 3.3.8

https://www.ruby-lang.org/en/news/2024/10/28/redos-rexml-cve-2024-49761/


Gemfile.lock
```diff
@@ -184,19 +186,20 @@ GEM
       mime-types (>= 1.16, < 4.0)
       netrc (~> 0.8)
     retriable (3.1.2)
-    rexml (3.3.8)
```
krugerk, line 187: https://www.ruby-lang.org/en/news/2024/10/28/redos-rexml-cve-2024-49761/

.github/workflows/fastlane-tests.yml
```diff
@@ -21,7 +21,7 @@ jobs:
     - name: Setup ruby and install gems
       uses: ruby/setup-ruby@v1
       with:
-        ruby-version: '2.7.2'
+        ruby-version: '3.3.6'
```
theospears, line 24: I'm guessing this is the cause of the fastlane failures?

theospears:
When I attempt to run this locally, I see the same infinite stack trace which was showing up in github actions:

```
> ruby --version
ruby 3.3.6 (2024-11-05 revision 75015d4c1f) [arm64-darwin24]
```

krugerk:
> When I attempt to run this locally, I see the same infinite stack trace which was showing up in github actions:
> 
> ```
> > ruby --version
> ruby 3.3.6 (2024-11-05 revision 75015d4c1f) [arm64-darwin24]
> ```
> 
> It looks like you worked around this for github actions via downgrade, but this is a pain locally as brew doesn't carry older versions, so we should probably hold off until this is fixed with latest ruby?

Locally I have rbenv where I have set ruby-version 3.3.5 for this repository.

theospears:
It looks like this works with ruby 3.3.6 if we upgrade to the latest version of bundler (see https://github.com/orgs/community/discussions/143096) so maybe let's do that?

theospears (approved):

Issue #450, opened 2024-04-09
Data entry should treat deadline as end-of-day


dreeves:
```[tasklist]
### Desiderata
- [x] Someone with an iPhone (dis?)confirms the replicata
```

## Replicata

1. Note the current time; say it's 11am on April 9
2. Make a test goal with deadline of 5 or 10 minutes from now, say 11:10am
3. Immediately add a datapoint of 123 for today
4. See that it shows up as 123 on April 9
5. Wait till the deadline passes, say it's 11:15am
6. Add a datapoint of 456 for today

## Expectata

That I get a datapoint for 456 on April 10. That's what happens on the website and in Beedroid.

## Resultata

I get a datapoint of 456 on April 9?

## Cognata

* https://doc.beeminder.com/midnight (in particular the appendix rationalizing this design choice)

_**Verbata**: beeminder day vs calendar day, deadline as end-of-day,_

dreeves:
Turns out this is already consistent with the web and Beedroid. Phew!

Issue #35, opened 2020-04-13
Bug: beemiOS uses calendar days, not Beeminder days


dreeves:
From Garrison Taylor:

> iOS bug that has been bothering me: if your goals are due at (say) 5am, on the Website it has the date automatically count as the previous day until that 5am deadline, while on the app, it counts as the current calendar day.

andrewpbrett:
dupe of #29, which is closed and will go out in the next release

PR #510, opened 2024-11-08
using ``Daystamp`` when considering deadline

### LINKED ISSUE #17: Built-in timer doesn't respect Beeminder definition of end-of-day
User reports submitting a timer at 2 minutes after a 13:00 deadline. The actual calendar day is the 26th, user expects datapoint with date of the 27th, since that's how Beeminder deadlines behave, but gets a datapoint with date of the 26th.

(I checked the server logs and confirmed that the Add Datapoint request was received at 2 minutes after the deadline, and the datapoint was submitted with `urtext: 26 <value> <comment>`.)

## Cognata

* #35 
* #450
* https://doc.beeminder.com/midnight (in particular the appendix rationalizing this design choice)

_**Verbata**: timer mode, beeminder day vs calendar day, end-of-day definition, deadline as end-of-day,_

### LINKED ISSUE #228: Datapoints entered after midnight don’t count towards correct day for certain deadlines
```[tasklist]
### Desiderata
- [ ] Replicata
- [ ] TBD
```

Report:
> I have some goals with a deadline at 00:50 am. Whenever I enter a datapoint between midnight and this deadline, the datapoint will already be counted toward the next day (technically the current day, but the next day from the point of view of this goal - hard to explain but I hope you know what I mean). Therefore, I always have to dial the date of last-minute datapoints back by one day manually (otherwise I would derail unjustifiedly). I have not seen this happening with goals where the deadline is 3 am or 4 pm or any other time, so I am guessing it has something to do with the fact that the deadline is between midnight and 1 am.

Place where the code maybe tries to address this (HT krugerk):
https://github.com/beeminder/BeeSwift/blob/e0db275bffd9cc51694f3df28244fd9f49c2c05f/BeeSwift/TimerViewController.swift#L138

## Cognata

* #26 
* #361 
* #248 

_**Verbata**: calendar day vs beeminder day, end-of-day,_

krugerk:
``Daystamp`` contains the logic for calculating the date (all of year, month, and day) that should be reported as the daystamp of a newly created datapoint, including accounting for the goal's deadline.

This merge request uses Daystamp and this feature of it to replace the three instances of other areas in the app tackling this same calculation: in Today View extension, in Goal / add datapoint, and in Timer.

Furthermore, three places with their own implementations all suffered the same logic error of not accounting for certain deadlines correctly - in particular deadlines just after midnight (midnight until 1 in the morning):

```swift
 if self.goal.deadline > 0 && currentHour! < 6 && currentHour! < self.goal.deadline/3600 {
            self.dateStepper.value = -1
        }
```
If there was not a whole hour of difference, then the deadline was effectively ignored. 

Fixes #17 
Fixes #228 

--
There have been many reports of the app not handling deadlines well or properly. Commit 5f3278c6f9a2d6fef21d63d7288739d31c81d1d3 introduces logic for handling the deadline in the Goal/add and Timer screens.
Other, possibly related issues: #17, #35, #228, #450

### DIFF
--- a/BeeSwift/GoalViewController.swift
+++ b/BeeSwift/GoalViewController.swift
@@ -209,30 +209,15 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
         self.dateStepper.tintColor = UIColor.Beeminder.gray
         dataEntryView.addSubview(self.dateStepper)
         self.dateStepper.addTarget(self, action: #selector(GoalViewController.dateStepperValueChanged), for: .valueChanged)
-        self.dateStepper.value = 0
-
-        // if the goal's deadline is after midnight, and it's after midnight,
-        // but before the deadline,
-        // default to entering data for the "previous" day.
-        let now = Date()
-        let calendar = Calendar.current
-        let components = (calendar as NSCalendar).components([.hour, .minute], from: now)
-        let currentHour = components.hour
-        if self.goal.deadline > 0 && currentHour! < 6 && currentHour! < self.goal.deadline/3600 {
-            self.dateStepper.value = -1
-        }
-
-        // if the goal's deadline is before midnight and has already passed for this calendar day, default to entering data for the "next" day
-        if self.goal.deadline < 0 {
-            let deadlineSecondsAfterMidnight = 24*3600 + self.goal.deadline
-            let deadlineHour = deadlineSecondsAfterMidnight/3600
-            let deadlineMinute = (deadlineSecondsAfterMidnight % 3600)/60
-            let currentMinute = components.minute
-            if deadlineHour < currentHour! ||
-                (deadlineHour == currentHour! && deadlineMinute < currentMinute!) {
-                self.dateStepper.value = 1
-            }
-        }
+        self.dateStepper.value = {
+            let now = Date()
+            let daystampAccountingForTheGoalsDeadline = Daystamp(fromDate: now,
+                                                                 deadline: goal.deadline)
+            let daystampAssumingMidnightDeadline = Daystamp(fromDate: now,
+                                                            deadline: 0)
+            
+            return Double(daystampAccountingForTheGoalsDeadline.distance(to: daystampAssumingMidnightDeadline))
+        }()
 
         self.dateStepper.snp.makeConstraints { (make) -> Void in
             make.top.equalTo(self.dateTextField.snp.bottom).offset(elementSpacing)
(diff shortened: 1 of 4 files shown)


BeeSwift/GoalViewController.swift
```diff
@@ -209,30 +209,15 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
...
-            let deadlineSecondsAfterMidnight = 24*3600 + self.goal.deadline
-            let deadlineHour = deadlineSecondsAfterMidnight/3600
-            let deadlineMinute = (deadlineSecondsAfterMidnight % 3600)/60
-            let currentMinute = components.minute
-            if deadlineHour < currentHour! ||
-                (deadlineHour == currentHour! && deadlineMinute < currentMinute!) {
-                self.dateStepper.value = 1
-            }
-        }
+        self.dateStepper.value = {
```
theospears, line 212: I agree this is an improvement and seems reasonable to merge. Long term we should probably change how the dateStepper works to use a representation more tied to daystamps and less to the current clock time.

BeeSwift/TimerViewController.swift
```diff
@@ -143,33 +143,11 @@ class TimerViewController: UIViewController {
...
-        }
-        
-        let formatter = DateFormatter()
-        formatter.locale = Locale(identifier: "en_US")
-        formatter.dateFormat = "d"
+        let urtextDaystamp: String = {
+            let daystamp = Daystamp(fromDate: Date(), deadline: goal.deadline)
+            
+            return String(format: "%04d %02d %02d", daystamp.year, daystamp.month, daystamp.day)
+        }()
```
theospears, line 150: Nit: For small units like this, I prefer not to use nested blocks. I think the benefits of nested scope are outweighed by the cost of having the expression that directly controls the variable value further from where it is declared.

In other words, this could just be: 
```
let daystamp = Daystamp(fromDate: Date(), deadline: goal.deadline)
let urtextDaystamp =  String(format: "%04d %02d %02d", daystamp.year, daystamp.month, daystamp.day)
```

theospears (approved):

BeeSwift/TimerViewController.swift
krugerk: I adjusted the code. What do you think?

theospears (approved):

BeeSwift/GoalViewController.swift
```diff
@@ -513,6 +490,15 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
...
         return self.goalImageView
     }
+    
+    private static func makeInitialDateStepperValue(date: Date = Date(), for goal: Goal) -> Double {
+        let daystampAccountingForTheGoalsDeadline = Daystamp(fromDate: date,
+                                                             deadline: goal.deadline)
+        let daystampAssumingMidnightDeadline = Daystamp(fromDate: date,
+                                                        deadline: 0)
+        
+        return Double(daystampAccountingForTheGoalsDeadline.distance(to: daystampAssumingMidnightDeadline))
```
krugerk, line 500: see #528 528

PR #513, opened 2024-11-12
using alamofireimage for transitions when loading thumbnails and graphs

krugerk:
It was already being used in the Today Extension view but not for the graphs and graph thumbnails elsewhere in the app. Using it simplifies the code and allows for easily adding transitions. When scrolling the gallery or loading a goal, the user experience is enhanced by using smooth rather than jarring animations and transitions.

### DIFF
--- a/BeeSwift/Components/GoalImageView.swift
+++ b/BeeSwift/Components/GoalImageView.swift
@@ -9,17 +9,18 @@ import BeeKit
 /// Shows the current graph for a goal
 /// Handles placeholders for loading and queued states, and automatically updates when the goal changes
 class GoalImageView : UIView {
-    private static let downloader = ImageDownloader(imageCache: AutoPurgingImageCache())
     private let logger = Logger(subsystem: "com.beeminder.com", category: "GoalImageView")
 
     private let imageView = UIImageView()
     private let beeLemniscateView = BeeLemniscateView()
 
     private var currentlyShowingGraph = false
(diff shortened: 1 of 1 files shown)


BeeSwift/Components/GoalImageView.swift
```diff
@@ -68,78 +69,40 @@ class GoalImageView : UIView {
...
-                self.showGraphImage(image: image)
-                break;
-            case .failure(let error):
-                self.logger.error("Error downloading goal graph: \(error)")
-                self.clearGoalGraph()
-            }
-        })
+        beeLemniscateView.isHidden = !currentlyShowingGraph
+        
+        imageView.af.setImage(withURLRequest: request,
```
krugerk, line 100: And here where it is being used in the Today View extension: https://github.com/krugerk/BeeSwift/blob/0456597a73674957ebd7734a7518cf6625f57d25/BeeSwiftToday/TodayTableViewCell.swift#L190

theospears:
I think this changes some current behavior which was intentional. In particular, the intended behavior is:
1. The bee lemniscate only appears while beebrain is active (i.e. the goal is being recalculated in the backend asynchronously), it's not intended to show while images are loading. This matches the behavior of the web app.
2. If we already have a graph image for a goal, but then get a new url from the server, don't show a loading screen, instead show the old graph until the new one is downloaded and ready.

theospears:
Going to close this out given the above, but comment here if you don't think that's the right action.

PR #514, opened 2024-11-15
Migrate from CoreData to SwiftData

theospears:
This migrates BeeSwift from CoreData to SwiftData. Data is not migrated, but this is fine as we have not done a public release based on CoreData. This is intended to be a minimal migration. In particular it does not yet make use of `@ModelActor` which will be needed to get the concurrency behavior right.

Note this removes Spotlight support as there isn't yet a handy helper for that with SwiftData like there is with core data. We'll need to write a little bit of manual code to bring that back.

Testing:
Verified the app launches, logs in, and can view gallery and goal page


theospears:
* **#514** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/514?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/514?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`

<h2></h2>

theospears:
This is the first working version. I need to read over the diff and also locally do some more validation that it isn't totally broken before merging. I intend to merge the smallest diff that basically works here to minimize the merge conflict window, and then follow up with more cleanup afterwards.

theospears:
Overall it seems like SwiftData isn't mature enough to use yet (see e.g. https://developer.apple.com/forums/thread/759364).

PR #527, opened 2024-11-21
HealthKitConfig.metrics

krugerk:
HealthKitConfig was meant to be a singleton, providing access to the list of supported HealthKit metrics. 
Used the `enum` for namespace with _static_ computed property for access to the goods.

### DIFF
--- a/BeeKit/HeathKit/HealthKitConfig.swift
+++ b/BeeKit/HeathKit/HealthKitConfig.swift
@@ -9,12 +9,9 @@
 import Foundation
 import HealthKit
 
-
-public class HealthKitConfig : NSObject {
-    public static let shared = HealthKitConfig()
-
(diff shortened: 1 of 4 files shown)


BeeKit/HeathKit/HealthKitConfig.swift
```diff
@@ -58,13 +55,8 @@ public class HealthKitConfig : NSObject {
             TimeInBedHealthKitMetric(humanText: "Time in bed", databaseString: "timeInBed", category: .Sleep),
             TimeAsleepHealthKitMetric(humanText: "Time asleep", databaseString: "timeAsleep", category: .Sleep),
 
-        ]
-
-        if #available(iOS 17, *) {
```
krugerk, line 63: bumped to min iOS 17 with #484

Issue #473, opened 2024-09-02
Date is wrong when app opened for the first time?


dreeves:
```[tasklist]
### Desiderata
- [ ] Replicata
```

Reported by user smokeskin:

> bug on the iPhone: when I open it the first time, the date always shows yesterday, but when I click forward it jumps forward 2 days to tomorrow. I assume that behind the scenes the date is set to today, but the frontend display isn’t updated. I can see it might be technically difficult to get the fronted to update, but the date forward button could perhaps be made to work relative to what the user sees?

_**Verbata**: date confusion, date staleness,_

theospears:
I think the user is talking about the date field in the add datapoint section at the bottom of the goal view?

krugerk:
I had what I thought was a similar issue: I tapped a notification, landing in the app. Expecting to be taken to the corresponding goal, when the goal view controller was shown, I immediately tapped submit, intending to +1 the goal corresponding to the alert.
I realized another goal was the one open and, upon further inspection, noticed that the day/daystamp of the just then added datapoint was for a few days ago. It must have been that that goal screen had been left open two days and the input date still had the data from then.
In my usual flow, when landing at that screen, I would have just come from the gallery/list of goals, and the date / value / comment fields would have all been set to fresh defaults.

krugerk:
As noted in the comments, it was not clear what was meant with the report. At the same time, a possible scenario could have been how the add data point section of the goal screen calculated the initial day to use. If it were that, #510 may have fixed it.

PR #528, opened 2024-11-23
initial date in add datapoint started with tomorrow rather than yesterday

krugerk:
Given the scenario, now is 1AM and one is entering into GoalVC for a goal with a deadline of 3AM the initial date shown was tomorrow. Expected might have been today and better yet is probably yesterday since 1AM is before the 3AM deadline, thus the daystamp for a datapoint added _now_ is still the (calendar) day before.

Fixes: #473


krugerk:
Fixes a bug introduced in #510.

theospears (approved):

theospears:
I've merged this, but could you open a follow up PR adding one or more regression tests around the date logic?

krugerk:
> I've merged this, but could you open a follow up PR adding one or more regression tests around the date logic?

draft here #529

PR #529, opened 2024-11-25
test of goalvc / add data initial stepper value

krugerk:
In order to test the logic for calculating the date initially shown in the add data section of a goal, a small _goalVM_ has been introduced for the goalVC.

https://github.com/beeminder/BeeSwift/pull/528#issuecomment-2495798319


BeeSwift/GoalViewModel.swift
```diff
@@ -0,0 +1,24 @@
...
+//
+//  Created by krugerk on 2024-11-25.
+//
+
+import Foundation
+import Intents
+
+import BeeKit
+
+struct GoalViewModel {
```
theospears, line 13: I like having somewhere for view-specific logic for goals to live which isn't in the model, and isn't in the UI classes. I don't think we've reached a point where it makes sense for that to be the main object we pass around though. 
1. It adds unnecessary indirection
2. We might end up wanting different "view models" for different views.

BeeSwiftTests/GoalViewModelTests.swift
```diff
@@ -0,0 +1,86 @@
+//
+//  GoalViewModelTests.swift
+//  BeeSwiftTests
+//
+//  Created by krugerk on 2024-11-25.
+//
```
theospears, line 6: I've been defaulting to just completely leaving out these header blocks. It's all in the git metadata anyway. Probably there's a project setting I should go change to change the default.

BeeSwiftTests/GoalViewModelTests.swift
```diff
@@ -0,0 +1,86 @@
...
+
+import Testing
+
+@testable import BeeSwift
+@testable import BeeKit
+
+struct GoalViewModelTests {
+    
+    @Test func initialStepperIsMinusOneWhenSubmissionDateIsAfterMidnightAndBeforeDeadline() async throws {
```
theospears, line 15: This is the new SwiftTesting syntax right? 😍

BeeSwiftTests/GoalViewModelTests.swift
```diff
@@ -0,0 +1,86 @@
...
+    
+    @Test func initialStepperIsZeroWhenSubmissionDateIsBeforeMidnightAndBeforeDeadline() async throws {
+        let goal = Self.makeGoalWithDeadline(0)
+        let viewModel = GoalViewModel(goal: goal)
+        let submissionDate = Calendar.current.date(bySettingHour: 20, minute: 30, second: 0, of: Date())!
+        let actual = viewModel.initialDateStepperValue(date: submissionDate)
+        #expect(actual == 0)
+    }
+
+    @Test func initialStepperIsPlusOneWhenSubmissionDateIsAfterDeadline() async throws {
```
theospears, line 31: Can we give the reader some context on _why_ PlusOne is the correct value in this context? Either via the name, or comments.

BeeSwiftTests/GoalViewModelTests.swift
```diff
@@ -0,0 +1,86 @@
...
+//
+//  Created by krugerk on 2024-11-25.
+//
+
+import Testing
+
+@testable import BeeSwift
+@testable import BeeKit
+
+struct GoalViewModelTests {
```
theospears, line 13: Thanks for writing tests to cover each of these cases 🎉

BeeSwiftTests/GoalViewModelTests.swift
```diff
@@ -0,0 +1,86 @@
...
+        let actual = viewModel.initialDateStepperValue(date: submissionDate)
+        #expect(actual == 1)
+    }
+}
+
+
+
+private extension GoalViewModelTests {
+    static func makeGoalWithDeadline(_ deadline: Int) -> Goal {
+        let context = BeeminderPersistentContainer.createMemoryBackedForTests().newBackgroundContext()
```
theospears, line 44: I wonder how CoreData contexts are meant to work in tests

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -489,7 +489,8 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
     }
     
     func openGoal(_ goal: Goal) {
-        let goalViewController = GoalViewController(goal: goal)
+        let viewModel = GoalViewModel(goal: goal)
```
theospears, line 492: Let's continue passing the `Goal` into `GoalViewController` and have it construct the `GoalViewModel` (see comment below for my reasons why)

BeeSwift/GoalViewModel.swift
krugerk: There is one approach where each and every view has its own view model.

BeeSwiftTests/GoalViewModelTests.swift
krugerk: In order for the correct value to appear in the view, some place has to know 'submission date aka now' and a goal's deadline. As is, this means knowing everything about a goal and thus about core data.

If the core data part could be more limited - maybe like in the mr for widgets #widgets - where there are conversions from nsmanagedobject goal to a smaller entity, containing just the parts that screen (widget) needs, then test setup and testing pieces should be easier and more focused.

BeeSwift/Gallery/GalleryViewController.swift
krugerk: Though we should move away from passing around nsmanagedobjects directly anyway. I did not change that here as we see.

BeeSwiftTests/GoalViewModelTests.swift
krugerk: and more such migration #530

theospears:
What's your plan here? I still think we should add tests for this logic, using old-style tests if necessary.

krugerk:
> What's your plan here? I still think we should add tests for this logic, using old-style tests if necessary.

Ah, I was cleaning up branches when I deleted this one which in turn closed this draft PR.

PR #530, opened 2024-11-26
migrate to Swift Testing

krugerk:
migrated from XCTest to Testing ... although it appears [fastlane does not pick up swift-testing tests](https://github.com/fastlane/fastlane/issues/22088)


theospears:
Out of an abundance of caution, could you prove these tests are running? e.g. I don't see them named in the outputs under Fastlane like they were previously.

krugerk:
> Out of an abundance of caution, could you prove these tests are running? e.g. I don't see them named in the outputs under Fastlane like they were previously.

I will look into it. Xcode runs them but yes I too see none of them being picked up by scan aka run_tests.

krugerk:
> Out of an abundance of caution, could you prove these tests are running? e.g. I don't see them named in the outputs under Fastlane like they were previously.

https://github.com/fastlane/fastlane/issues/22088

Issue #206, opened 2020-09-28
Use consistent color scheme based on safety buffer (safebuf)


krugerk:
[Probably make a new gissue pointing to https://github.com/beeminder/beeminder/issues/1755]

BeemiOS currently does not place its own dots. However, it does use different colors based on safety buffer.

See https://github.com/beeminder/road/issues/111

```
color = (safebuf < 1 ? "#ff0000" :  # Red for beemergencies
         safebuf < 2 ? "#ffa500" :  # Orange for 1 safe day
         safebuf < 3 ? "#3f3fff" :  # Blue for 2 safe days
         safebuf < 7 ? "#00aa00" :  # Green for 3+ safe days
                       "#228B22")   # Forest green Grayson dots for 7+ safe days
```

Here is the app using red, orange, blue, and green (and possibly gray) for the color of the countdown:
https://github.com/beeminder/BeeSwift/blob/31bfceaeaa126adc9290b45afb17fc08832c8279/BeeSwift/JSONGoal.swift#L170-L182

And here the app uses green, blue, and orange as its delta colors:
https://github.com/beeminder/BeeSwift/blob/31bfceaeaa126adc9290b45afb17fc08832c8279/BeeSwift/JSONGoal.swift#L279-L291

## Cognata

* #55

_**Verbata**: red, orange, blue, green, dot colors, color scheme,_

krugerk:
Keep #55 in mind

dreeves:
Ah, interesting. But I'm confused about what needs to change on the BeemiOS / Swift side. Where do things like `UIColor.red` come from?

krugerk:
[UIColor](https://developer.apple.com/documentation/uikit/uicolor) comes from Apple's UIKit. Red and a bunch of other standard colors have already been defined. The app does not show its own dots, just the graph it downloads. However, it does use color based on safety buffer for the amounts due ("+5" in blue, "+15" in green) and for a textual representation of the safety buffer ("Safe for 3 days")

dreeves:
Got it. So this change can't be effected server-side, right? We need BeemiOS to implement the above function mapping number of safe days (`safebuf` in the API) to the actual color hex code?

krugerk:
Correct. Originally I posted that the app obtains colors from the backend. Not true. It just uses safety buffer, part of a goal, provided by the backend, to determine colors.

krugerk:
[...]

PR #535, opened 2024-12-02
Color safeSum in forest green as used in the Grayson Dot

krugerk:
## Summary
The chart includes the Grayson Dot - a different color green applied to green dots with more than 7d buffer. The iOS app colors the `safeSum` using colors based on the colors of the dots. This text appears in both the gallery in a goal's cell as well as in the goal screen itself. Previously the same green was being used for all `safeSum` with a `safeBuf` greater than three.

## Validation
Ran app in the simulator using the new colors and logic. Made screenshots.

grayson-green-for-safety-buffer-more-than-seven-on-dark

Implements #206

### DIFF
--- a/BeeKit/UI/UIColorExtension.swift
+++ b/BeeKit/UI/UIColorExtension.swift
@@ -27,5 +27,13 @@ extension UIColor {
                                            green: 217.0/255.0,
                                            blue: 17.0/255.0,
                                            alpha: 1)
+        
+        public struct SafetyBuffer {
+            public static let red: UIColor = .systemRed // .init(red: 1, green: 0, blue: 0, alpha: 1)
+            public static let orange: UIColor = .systemOrange // .init(red: 1, green: 165/255.0, blue: 00, alpha: 1)
+            public static let blue: UIColor = .systemBlue // .init(red: 63/255.0, green: 63/255.0, blue: 1, alpha: 1)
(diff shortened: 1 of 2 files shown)


BeeKit/UI/UIColorExtension.swift
```diff
@@ -12,20 +12,21 @@ import UIKit
...
-                                          green: 163.0/255.0,
-                                          blue: 81.0/255.0,
-                                          alpha: 1)
-        
-        public static let blue: UIColor = .systemBlue
-        public static let orange: UIColor = .systemOrange
         public static let red: UIColor = .systemRed
         
-        public static let gray = UIColor(white: 0.7, alpha: 1.0)
+        public static let gray = UIColor.systemGray
```
theospears, line 17: I believe this is a somewhat darker color than the previous one (more like `white: 0.5`). Was that deliberate?

theospears (commented):
The color parts of this look good. Less sure about the change to gray.

BeeKit/UI/UIColorExtension.swift
theospears: > The [system gray one will adapt to the device's environment](https://developer.apple.com/documentation/uikit/uicolor/3173143-systemgray)

I agree this is true in abstract, and for most system colors. It turns out `systemGray` is the same in both dark mode and light mode. I'm not convinced having this universal gray really makes sense (e.g. perhaps backgrounds should be closer to the background color than borders are), but that seems like a problem for another day.

PR #536, opened 2024-12-02
Avoid keyboard obscuring last goals in gallery view

theospears:
## Summary
Previously when using the goal filter, the keyboard would obscure the last goals in the list. Here we make the scrollview resize itself to only take up the section of screen above the keyboard so you can always scroll to the bottom.

This is an alternative approach to #503

## Validation
Loaded the gallery view on device
* Verified when keyboard is hidden the scroll view expands to take up the entire screen
* Verified when the filter box is focused and the keyboard appears it is still possible to scroll down to last goal

### DIFF
--- a/BeeSwift/Gallery/GalleryViewController.swift
+++ b/BeeSwift/Gallery/GalleryViewController.swift
@@ -147,7 +147,7 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
             make.top.equalTo(self.searchBar.snp.bottom)
             make.left.equalTo(self.view.safeAreaLayoutGuide.snp.leftMargin)
             make.right.equalTo(self.view.safeAreaLayoutGuide.snp.rightMargin)
-            make.bottom.equalTo(0)
+            make.bottom.equalTo(self.collectionView!.keyboardLayoutGuide.snp.top)
         }
         
         self.view.addSubview(self.noGoalsLabel)


BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -147,7 +147,7 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
             make.top.equalTo(self.searchBar.snp.bottom)
             make.left.equalTo(self.view.safeAreaLayoutGuide.snp.leftMargin)
             make.right.equalTo(self.view.safeAreaLayoutGuide.snp.rightMargin)
-            make.bottom.equalTo(0)
+            make.bottom.equalTo(self.collectionView!.keyboardLayoutGuide.snp.top)
```
krugerk, line 150: 👀 [keyboardLayoutGuide](https://developer.apple.com/documentation/uikit/uiview/3752221-keyboardlayoutguide)

looks great! 👏 
This one line using this iOS 15 feature sure beats out the manual 27 line approach in #503.

(_bothers me that we have another use of collectionView! - we could just mark it as such at Line 23 and not have to do it anywhere else instead of doing it everywhere else, though obviously this is not directly related to adjusting the view's bottom to handle the keyboard_)

krugerk:
This one then fixes #200

PR #537, opened 2024-12-03
Introduce a freshness indicator for the goal screen

### LINKED ISSUE #246: Loading indicator for goals
```[tasklist]
### Desiderata
- [ ] Does this replicate in the current version of the app?
- [ ] TBD
```

Request [from the forum](https://forum.beeminder.com/t/ios-app-loading-indicator-for-goals/5595):

> When I launch the iOS app, it often gives me a scare. If I’ve entered data for one of my goals outside the app, and that goal would have otherwise derailed, then it shows a big red LOST when the app first opens. It sends a stab of fear through me, until the app updates its data and refreshes to show no, my goal is still on track. Phew.
>
> This happens almost every time I open the app. Can you add a neutral “loading” indicator when it launches, so it doesn’t just assume I’ve derailed? Or maybe just for the goals that would have derailed, wait until you’ve actually refreshed the data before showing that scary red indicator.

The app (Version 5.8) does display a textual representation of the time the list of goals was last updated.

## Cognata

* https://github.com/beeminder/BeeSwift/blob/120beaaf96c06082280141f9acac09775a4d2091/BeeSwift/GalleryViewController.swift#L345

_**Verbata**: microcopy, status indicators, freshness / staleness of graphs, scaring poor users with spurious derailment indicators,_

krugerk:
## Summary
In line with #246, this provides an indicator of freshness on the goal screen. It reuses the one from the gallery screen.

## Media

https://github.com/user-attachments/assets/b2549822-e1b0-42de-9dea-f451bee4fad4

## Validation
Ran app in simulator
Viewed both gallery and goal, navigating back and forth.
Sent the app to background, back into foreground, varying sometimes with the gallery open, sometimes with a goal open.
Also watched the last updated since label be updated while just leaving the app open with the goal view showing

Fixes #246

### DIFF
--- a/BeeKit/Managers/GoalManager.swift
+++ b/BeeKit/Managers/GoalManager.swift
@@ -25,9 +25,6 @@ public actor GoalManager {
     private nonisolated let currentUserManager: CurrentUserManager
 
     public var goalsFetchedAt : Date? = nil
-    
-    // id, not goalname (slug)
-    private var lastFetchedByGoalId: [String: Date?] = [:]
 
     private var queuedGoalsBackgroundTaskRunning : Bool = false
 
@@ -68,9 +65,6 @@ public actor GoalManager {
         self.updateGoalsFromJson(response) // TODO: Return failure info
 
         self.goalsFetchedAt = Date()
-        staleGoals(context: modelContext)?.map(\.id).forEach { goalId in
-            lastFetchedByGoalId.updateValue(self.goalsFetchedAt, forKey: goalId)
-        }
 
         await performPostGoalUpdateBookkeeping()
     }
@@ -86,22 +80,12 @@ public actor GoalManager {
         goal.updateToMatch(json: goalJSON)
 
         try modelContext.save()
-        
-        lastFetchedByGoalId.updateValue(Date(), forKey: goal.id)
 
         await performPostGoalUpdateBookkeeping()
     }
 
     public func forceAutodataRefresh(_ goal: Goal) async throws {
         let _ = try await requestManager.get(url: "/api/v1/users/\(currentUserManager.username!)/goals/\(goal.slug)/refresh_graph.json", parameters: nil)
-    }
-    
-    
-    /// when a goal was last fetched
-    /// - Parameter goalId: goal.id of interest
-    /// - Returns: Date a goal was last fetched or nil
-    public func getDateLastFetchedForGoalWithId(_ goalId: String) -> Date? {
-        lastFetchedByGoalId[goalId]?.map { $0 }
     }
 
     private func updateGoalsFromJson(_ responseJSON: JSON) {
@@ -208,7 +192,6 @@ public actor GoalManager {
 
     private func resetStateForSignOut() {
         self.goalsFetchedAt = Date(timeIntervalSince1970: 0)
-        self.lastFetchedByGoalId = [:]
 
         // TODO: Delete from CoreData
     }
--- a/BeeKit/Managers/GoalManager.swift
+++ b/BeeKit/Managers/GoalManager.swift
@@ -27,7 +27,7 @@ public actor GoalManager {
     public var goalsFetchedAt : Date? = nil
     
     // id, not goalname (slug)
-    public private(set) var lastFetchedByGoalId: [String: Date?] = [:]
+    private var lastFetchedByGoalId: [String: Date?] = [:]
 
     private var queuedGoalsBackgroundTaskRunning : Bool = false
 
@@ -94,6 +94,14 @@ public actor GoalManager {
 
     public func forceAutodataRefresh(_ goal: Goal) async throws {
         let _ = try await requestManager.get(url: "/api/v1/users/\(currentUserManager.username!)/goals/\(goal.slug)/refresh_graph.json", parameters: nil)
+    }
+    
+    
+    /// when a goal was last fetched
+    /// - Parameter goalId: goal.id of interest
+    /// - Returns: Date a goal was last fetched or nil
+    public func getDateLastFetchedForGoalWithId(_ goalId: String) -> Date? {
+        lastFetchedByGoalId[goalId]?.map { $0 }
     }
 
     private func updateGoalsFromJson(_ responseJSON: JSON) {
--- a/BeeKit/Managers/GoalManager.swift
+++ b/BeeKit/Managers/GoalManager.swift
@@ -72,7 +72,7 @@ public actor GoalManager {
     public func refreshGoal(_ goalID: NSManagedObjectID) async throws {
         let goal = try modelContext.existingObject(with: goalID) as! Goal
 
-        let responseObject = try await requestManager.get(url: "/api/v1/users/\(goal.owner.username)/goals/\(goal.slug)?datapoints_count=5", parameters: nil)
+        let responseObject = try await requestManager.get(url: "/api/v1/users/\(currentUserManager.username!)/goals/\(goal.slug)?datapoints_count=5", parameters: nil)
         let goalJSON = JSON(responseObject!)
 
         // The goal may have changed during the network operation, reload latest version
@@ -85,7 +85,7 @@ public actor GoalManager {
     }
 
     public func forceAutodataRefresh(_ goal: Goal) async throws {
-        let _ = try await requestManager.get(url: "/api/v1/users/\(goal.owner.username)/goals/\(goal.slug)/refresh_graph.json", parameters: nil)
+        let _ = try await requestManager.get(url: "/api/v1/users/\(currentUserManager.username!)/goals/\(goal.slug)/refresh_graph.json", parameters: nil)
     }
 
     private func updateGoalsFromJson(_ responseJSON: JSON) {
--- a/BeeKit/Managers/GoalManager.swift
+++ b/BeeKit/Managers/GoalManager.swift
@@ -25,6 +25,9 @@ public actor GoalManager {
     private nonisolated let currentUserManager: CurrentUserManager
 
     public var goalsFetchedAt : Date? = nil
+    
+    // id, not goalname (slug)
+    public private(set) var lastFetchedByGoalId: [String: Date?] = [:]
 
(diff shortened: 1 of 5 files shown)


BeeKit/Managers/GoalManager.swift
```diff
@@ -25,6 +25,9 @@ public actor GoalManager {
     private nonisolated let currentUserManager: CurrentUserManager
 
     public var goalsFetchedAt : Date? = nil
+    
+    // id, not goalname (slug)
+    private var lastFetchedByGoalId: [String: Date?] = [:]
```
theospears, line 30: I think `Goal.lastModifiedLocal` may already be tracking the value you want?

krugerk: I was looking for when the app last fetched the data from the server. With Core Data, theoretically we could be editing objects locally (offline mode?) and at some point syncing them with the server.

Given _local_ in the name and the description of the attribute,
https://github.com/beeminder/BeeSwift/blob/745cf9875ed6b0303c46d5d8e4c0a2d44796040d/BeeKit/Model/Goal.swift#L66-L67
I had the impression this is meant to reflect when the goal was last touched locally and does not necessarily reflect also when the goal was last fetched from the server. I see this date is being set whenever updateToMatch is being called which is being called after a fetch of user/goal from the backend.

krugerk: I see it, `goal.lastModifiedLocal` is also set with Date.now when the object is created (pulled from the db). In this case the lastModifiedLocal does not reflect how fresh the data is (when last pulled from server).

theospears: > goal.lastModifiedLocal is also set with Date.now when the object is created (pulled from the db). 

I don't believe this is true, at the very least it's not the intended behavior. While there is a constructor which sets the value, I don't believe this is the constructor which is invoked when materializing from the database.

krugerk: Alright, you convinced me!

theospears: Yeah, I thought the explicit list all attributes constructors might be useful for tests. If we had more tests.

BeeKit/Managers/GoalManager.swift
```diff
@@ -72,19 +72,20 @@ public actor GoalManager {
     public func refreshGoal(_ goalID: NSManagedObjectID) async throws {
         let goal = try modelContext.existingObject(with: goalID) as! Goal
 
-        let responseObject = try await requestManager.get(url: "/api/v1/users/\(currentUserManager.username!)/goals/\(goal.slug)?datapoints_count=5", parameters: nil)
+        let responseObject = try await requestManager.get(url: "/api/v1/users/\(goal.owner.username)/goals/\(goal.slug)?datapoints_count=5", parameters: nil)
```
theospears, line 75: This change seems fine and we can merge with it in, though I'll once again request to minimize unrelated cleanup (there are no feature related changes to this file) in feature PRs.

BeeKit/Managers/RequestManager.swift
```diff
@@ -39,6 +39,8 @@ public class RequestManager {
             urlWithSubstitutions = urlWithSubstitutions.replacingOccurrences(of: "{username}", with: username)
         }
 
+        logger.debug("\(#function) | \(method.rawValue) | \(urlWithSubstitutions)")
```
theospears, line 42: Again, unrelated change.

As a reviewer it can be hard to know if changes like this are deliberate, or were intended as temporary debugging measures, and weren't cleaned up before opening the PR.

BeeSwift/GoalViewController.swift
```diff
@@ -24,6 +24,14 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
 
     let goal: Goal
 
+
+    private var lastFetched: Date? {
```
theospears, line 28: This doesn't seem to be a real variable. By this I mean we don't really keep track of a value in it - the only reads are within `updateLastUpdatedLabel`, which could just as happily be reading directly from `goal.lastModifiedLocal`

[Do] Let's remove it, and just call `updateLastUpdatedLabel()` in the places we were writing to it.

BeeSwift/GoalViewController.swift
```diff
@@ -63,9 +72,20 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
...
 
+        self.view.addSubview(self.timeElapsedView)
+        self.timeElapsedView.snp.makeConstraints { (make) -> Void in
+            make.top.equalTo(self.view.safeAreaLayoutGuide.snp.topMargin)
+            make.left.equalTo(0)
+            make.right.equalTo(0)
+        }
+        
+        self.updateLastUpdatedLabel()
+        Timer.scheduledTimer(timeInterval: 60, target: self, selector: #selector(GoalViewController.updateLastUpdatedLabel), userInfo: nil, repeats: true)
```
theospears, line 83: [Consider] Wouldn't this timer make more sense being part of the `FreshnessIndicatorView` ?

theospears (changes requested):
Overall approach looks good, and I like the idea of extracting `FreshnessIndicatorView` as its own component. I added some requests for local changes.

When testing, could you also verify the time since update changes when you just leave the app open with the goal view showing?

BeeKit/Managers/RequestManager.swift
krugerk: It is not necessary for the MR itself though it was deliberate to facilitate development.

I am of the opinion that one should be able to follow log statements back to figure out what went wrong. The logging in the app is not yet to that point.

theospears (changes requested):
Still waiting on changes to GoalViewController

BeeSwift/GoalViewController.swift
krugerk: The reason I placed it here is because this is also a Controller whereas the Freshness one is just a view.

BeeKit/Managers/GoalManager.swift
krugerk: removed from this PR

BeeSwift/GoalViewController.swift
```diff
@@ -24,6 +24,14 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
...
+    private var lastFetched: Date? {
+        didSet {
+            updateLastUpdatedLabel()
+        }
+    }
+    
+    private let timeElapsedView = FreshnessIndicatorView()
     fileprivate var goalImageView = GoalImageView(isThumbnail: false)
     fileprivate var datapointTableController = DatapointTableViewController()
     fileprivate var dateTextField = UITextField()
```
krugerk, line 30:
```suggestion
    fileprivate let goalImageView = GoalImageView(isThumbnail: false)
    fileprivate let datapointTableController = DatapointTableViewController()
    fileprivate let dateTextField = UITextField()
```

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -22,8 +22,7 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
 
     var collectionView :UICollectionView?
     var collectionViewLayout :UICollectionViewFlowLayout?
-    let lastUpdatedView = UIView()
-    let lastUpdatedLabel = BSLabel()
+    private let freshnessIndicator = FreshnessIndicatorView()
     let cellReuseIdentifier = "Cell"
     var deadbeatView = UIView()
     var outofdateView = UIView()
```
krugerk, line 28:
```suggestion
    private let deadbeatView = UIView()
    private let outofdateView = UIView()
```

BeeKit/Managers/RequestManager.swift
krugerk: removed from this PR

BeeSwift/GoalViewController.swift
krugerk: done

theospears (approved):

theospears:
LGTM, merged!

BeeSwift/GoalViewController.swift
krugerk: unfortunately that has the side effect of the app crashing when accessing goal.lastModifiedLocal at GoalViewController.updateLastUpdatedLabel upon signing into the app.

1. GoalManager.performPostGoalUpdateBookkeeping posts notification goalsUpdated
2. GoalViewController.onGoalsUpdatedNotification executes
3. calling its updateInterfaceToMatchGoal
4. GoalViewController.updateInterfaceToMatchGoal calls, among others, updateLastUpdatedLabel
5. GoalViewController.updateLastUpdateLabel now accesses self.goal.lastModifiedLocal , which at the time, is still part of an empty NSManagedObject Goal:

```swift
<Goal: 0x600002148aa0> (entity: Goal; id: 0xblahblahblah9494 <x-coredata://blahblah-F9F9-4FB3-BD1B-blahblah/Goal/p385>; data: {
    alertStart = nil;
    autodata = nil;
    data = "<relationship fault: 0x600000383740 'data'>";
    deadline = nil;
    graphUrl = nil;
    healthKitMetric = nil;
    hhmmFormat = nil;
    id = nil;
    initDay = nil;
    lastModifiedLocal = nil;
    lastTouch = nil;
    leadTime = nil;
    limSum = nil;
    owner = nil;
    pledge = nil;
    queued = nil;
    recentData = "<relationship fault: 0x6000003829a0 'recentData'>";
    safeBuf = nil;
    safeSum = nil;
    slug = nil;
    thumbUrl = nil;
    title = nil;
    todayta = nil;
    urgencyKey = nil;
    useDefaults = nil;
    won = nil;
    yAxis = nil;
})
```

theospears: It seems likely that versions of this bug will end up existing, or be easy to introduce, all across GoalView. Do you have ideas for why the GoalViewController even exists while in this state? We're certainly not showing it.

theospears: Tracking this as #590

PR #539, opened 2024-12-03
more zoom levels and more powerful zoom in on goal graph

krugerk:
## Summary
Via double tap on the goal's graph, step through the zoom levels.

One could double tap on the goal's graph to see a zoomed in version. This was at 2x magnification. Now it goes to 7x and from 1x to 3x to 5x and then 7x.

The graph (wrapped still inside a scrollview) still supports the pan to zoom. This zoom's max has also been bumped with this MR.

## Validation
Ran the app in the iPhone simulator. Tapped and zoomed on some graphs.

### DIFF
--- a/BeeSwift/GoalViewController.swift
+++ b/BeeSwift/GoalViewController.swift
@@ -363,7 +363,8 @@ class GoalViewController: UIViewController, DatapointTableViewControllerDelegate
 
     @objc func goalImageTapped() {
         let possibleNextStepUp = self.goalImageScrollView.zoomScale + 2.0
-        let nextZoomLevel: Double = possibleNextStepUp <= 7.0 ? possibleNextStepUp : 1.0
+        let nextZoomLevel = possibleNextStepUp <= self.goalImageScrollView.maximumZoomScale ? possibleNextStepUp : self.goalImageScrollView.minimumZoomScale
+        
         self.goalImageScrollView.setZoomScale(nextZoomLevel, animated: true)
     }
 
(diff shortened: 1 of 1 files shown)


BeeSwift/GoalViewController.swift
```diff
@@ -353,7 +353,10 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
     }
 
     @objc func goalImageTapped() {
-        self.goalImageScrollView.setZoomScale(self.goalImageScrollView.zoomScale == 1.0 ? 2.0 : 1.0, animated: true)
+        let possibleNextStepUp = self.goalImageScrollView.zoomScale + 2.0
+        let nextZoomLevel = possibleNextStepUp <= self.goalImageScrollView.maximumZoomScale ? possibleNextStepUp : self.goalImageScrollView.minimumZoomScale
+        
+        self.goalImageScrollView.setZoomScale(nextZoomLevel, animated: true)
```
theospears, line 359: This seems reasonable for iPhone. I'm less sure it makes sense for iPad though? Perhaps the max zoom level should be a function of the screen size?

krugerk: I still found it quite OK to double tap to zoom on the iPad. The bigger issue for me was that the graph image being used quickly becomes blurry since we are using the lower resolution png rather than the svg. The user can still pinch-to-zoom and can do so at the previous level as well as now also at other levels.

krugerk:
As for the Android app, double tapping the graph zooms in once, double tapping again zooms in yet further. Double tapping a third time zooms back out.

PR #540, opened 2024-12-04
introduces goal menu with deep links to goal on website

krugerk:
## Summary
The goal screen contained an action button which opened the goal in an in-app browser. This way one could manipulate various aspects of the goal or view data (statistics) from on the website as these features are lacking in the app.

This merge request replaces the action button on the goal screen, with single link to the goal page on the website, with a menu with links to all of the main sections of the website's rendition of the goal: commitment, stop/pause, data, statistics, and settings. This makes it easier to more quickly arrive at a particular section. It might also increase feature discoverability.

For example, the delta text was removed from the app and the current workaround to check "how much to do to earn x days off" is available on the website in the statistics section under the "Amounts Due By Day" subsection.
Also, the app does not support features such as setting the goal's description or title. It also does not allow editing of datapoints of goals with autodata whereas the website does.

## before
Tapping the action button opens beeminder.com/user/goal in an in-app browser after which one can navigate through the sections of the goal on the webpage presented.
before

#### example:
links from goal to goal sections on website - example
#### generic:
links from goal to goal sections on website - generic

## Validation
Opened the app in the simulator. Clicked through various goals and the corresponding "open this section of the goal on the website" links.

### DIFF
--- a/BeeSwift/GoalViewController.swift
+++ b/BeeSwift/GoalViewController.swift
@@ -288,69 +288,6 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
         updateInterfaceToMatchGoal()
     }
     
-    enum MenuAction {
-        case goalCommitment
-        case goalStop
-        case goalData
-        case goalStatistics
-        case goalSettings
-        
-        func makeLink(username: String, goalName: String) -> URL? {
-            guard
-                let accessToken = ServiceLocator.currentUserManager.accessToken
-            else { return nil }
-            
-            let destinationUrl: URL
-            
-            switch self {
-            case .goalCommitment:
-                destinationUrl = DeeplinkGenerator.generateDeepLinkToGoalCommitment(username: username, goalName: goalName)
-            case .goalStop:
-                destinationUrl = DeeplinkGenerator.generateDeepLinkToGoalStop(username: username, goalName: goalName)
-            case .goalData:
-                destinationUrl = DeeplinkGenerator.generateDeepLinkToGoalData(username: username, goalName: goalName)
-            case .goalStatistics:
-                destinationUrl = DeeplinkGenerator.generateDeepLinkToGoalStatistics(username: username, goalName: goalName)
-            case .goalSettings:
-                destinationUrl = DeeplinkGenerator.generateDeepLinkToGoalSettings(username: username, goalName: goalName)
-            }
-            
-            return DeeplinkGenerator.generateDeepLinkToUrl(accessToken: accessToken, username: username, url: destinationUrl)
-        }
-    }
-    
-    struct MenuOption {
-        let title: String
-        let action: MenuAction
-        let imageSystemName: String
-    }
-    
-    private func getMenuOptions() -> [MenuOption] {
-        [
-            MenuOption(title: "Commitment", action: .goalCommitment, imageSystemName: "signature"),
-            MenuOption(title: "Stop/Pause", action: .goalStop, imageSystemName: "pause.fill"),
-            MenuOption(title: "Data", action: .goalData, imageSystemName: "tablecells"),
-            MenuOption(title: "Statistics", action: .goalStatistics, imageSystemName: "chart.bar.fill"),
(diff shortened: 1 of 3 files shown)


BeeSwift/GoalViewController.swift
```diff
@@ -284,7 +287,70 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
         setValueTextField()
         updateInterfaceToMatchGoal()
     }
-
+
```
theospears, line 290: Is there somewhere outside `GoalViewController` it would make sense to extract this to?

BeeSwift/GoalViewController.swift
```diff
@@ -284,7 +287,70 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
...
+            MenuOption(title: "Statistics", action: .goalStatistics, imageSystemName: "chart.bar.fill"),
+            MenuOption(title: "Settings", action: .goalSettings, imageSystemName: "gearshape.2"),
+        ]
+    }
+    
+    private func createGoalMenu() -> UIMenu {
+        let options = getMenuOptions()
+        let actions = options.map { option in
+            UIAction(title: option.title, image: UIImage(systemName: option.imageSystemName), handler: { [weak self] _ in
+                guard let self else { return }
```
theospears, line 342: This is a fun little bit of syntax. 👍

BeeSwift/GoalViewController.swift
```diff
@@ -284,7 +287,70 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
...
+    struct MenuOption {
+        let title: String
+        let action: MenuAction
+        let imageSystemName: String
+    }
+    
+    private func getMenuOptions() -> [MenuOption] {
+        [
+            MenuOption(title: "Commitment", action: .goalCommitment, imageSystemName: "signature"),
+            MenuOption(title: "Stop/Pause", action: .goalStop, imageSystemName: "pause.fill"),
```
theospears, line 331: With this icon the button ends up looking more like an action (i.e. clicking it will stop/pause the goal) rather than a link to a page. Any ideas for an alternative?

theospears (commented):
Overall looks good! Added some small notes.

BeeSwift/GoalViewController.swift
krugerk: A few options come to mind:
- No icons
- all use the sf symbol safari since they all open in the browser
- introduce [TipKit](https://developer.apple.com/documentation/tipkit/highlightingappfeatureswithtipkit) to suggest using the feature (explaining a tad what they do)
- leave it as is
- use other icons
- reword title of menu, possibly of entries as well

BeeSwift/GoalViewController.swift
krugerk: ah it was very common for a while to declare weak self then avoid self? it each time by only executing the block when self was then non-nil

And instead of `guard let self = self else { return }` we may have seen `guard let strongSelf = self else...`.
Thanks to the syntax highlighting in Xcode we can usually see which one is which (local, strong self, vs local, weak self vs the entity's own self). And of course the compiler calculates it for us too.

BeeSwift/GoalViewController.swift
krugerk: My next move would be just to refactor it into a private extension of GoalVC further down in this file.

BeeSwift/GoalViewController.swift
theospears: I've thought about this some more, and decided:
1) I don't immediately have an idea I love
2) If/when people are confused about this, they'll click it, and learn what it actually does without harm.

theospears:
Could you update the screenshots to match the latest version?

theospears (approved):

krugerk:
A user had described (June 2023) some use cases of using the 'open goal on website in browser' link

> viewing more data than the app shows
> editing past data
> viewing the commitment dial (and changing it would be nice too)
> viewing the “amounts due by day” from the statistics tab.
> viewing the “take a break” section from the stop/pause tab (and changing it would be nice too)

https://forum.beeminder.com/t/ios-app-update/11022/6

PR #541, opened 2024-12-05
beeswift need only link to beekit once

krugerk:
## Summary
Sometimes, despite changes to something in BeeKit, Xcode would complain in BeeSwift source code that seemed to reflect the older state in BeeKit. Looking around for possible causes, it was noticed that target BeeSwift linked BeeKit twice. A merge artifact? One of the two was removed with this merge request.

## Validation
Built app in Xcode.
Launched in a Simulator.
Navigated in app, gallery, goals, settings, ...


theospears (approved):
LGTM. Thanks for tracking this down and fixing it!

PR #542, opened 2024-12-05
Switch the gallery view to use a StackView instead of manual layout

theospears:
## Summary
Previously we used SnapKit and lots of constraints in order to control the layout of the gallery view. This was a bunch of code, and cause constraint violation warnings due to margins being impossible in conjunction with hiding elements by setting their height to 0.

Here we migrate to use UIStackView which automatically handles constraints, sizing, and removing hidden views. This lets us delete a lot of code, and also fixes a bug where there would be a rendering glitch if returning to the gallery view from a goal view where the keyboard was hidden - previously the area the keyboard had occupied would appear blank.

Known issues to fix before merging: The background for last updated, deadbeat, etc, do not use the full width of the screen.

## Validation
Loaded the app on iPhone and iPad
* Checked the layout looks good in portrait and landscape
* Checked with and without keyboard
* Checked navigating from goal view back to gallery view
* Confirmed with the keyboard visible you can still scroll all goals

### DIFF
--- a/BeeSwift/Gallery/GalleryViewController.swift
+++ b/BeeSwift/Gallery/GalleryViewController.swift
@@ -20,6 +20,7 @@ import BeeKit
 class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayout, UICollectionViewDelegate, UICollectionViewDataSource, UISearchBarDelegate, SFSafariViewControllerDelegate {
     let logger = Logger(subsystem: "com.beeminder.beeminder", category: "GalleryViewController")
 
+    var stackView = UIStackView()
     var collectionView :UICollectionView?
     var collectionViewLayout :UICollectionViewFlowLayout?
     private let freshnessIndicator = FreshnessIndicatorView()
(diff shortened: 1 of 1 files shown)


theospears:
* **#542** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/542?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/542?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`

<h2></h2>

krugerk:
Yes Please! More UIStackView and less SnapKit. On our way toward SwiftUI.

Two more things for the list of known issues:
 - In the simulator, with the search bar present and the soft keyboard toggled on, tapping search off and back on, we see the last fetched and probably rest of the stack view coming in from the top of the device, passing through the safe area, passing through the navbar, on their way to being placed just under the navigation bar.
 - After signing in, the gallery is shown and initially has 0 goals while it fetches some. The 'you have no goals' message is shown quite low on the screen, much lower than it was before. Maybe it can be placed further up, after all, it is the only thing in the stack view, or in the middle, or we can re-check #506

krugerk:
> and also fixes a bug where there would be a rendering glitch if returning to the gallery view from a goal view where the keyboard was hidden - previously the area the keyboard had occupied would appear blank.

and in the meantime, #545

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -20,6 +20,8 @@ import BeeKit
 class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayout, UICollectionViewDelegate, UICollectionViewDataSource, UISearchBarDelegate, SFSafariViewControllerDelegate {
     let logger = Logger(subsystem: "com.beeminder.beeminder", category: "GalleryViewController")
 
+    var stackView = UIStackView()
+    var collectionContainer = UIView()
```
krugerk, line 24:
```suggestion
    let stackView = UIStackView()
    let collectionContainer = UIView()
```

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -56,15 +69,12 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
         let item = UIBarButtonItem(image: UIImage(systemName: "gearshape.fill"), style: UIBarButtonItem.Style.plain, target: self, action: #selector(self.settingsButtonPressed))
         self.navigationItem.rightBarButtonItem = item
         
-        self.view.addSubview(self.lastUpdatedView)
+        stackView.addArrangedSubview(self.lastUpdatedView)
```
krugerk, line 72: How about configuring each individual view up here and then afterwards, toward the bottom of this method, setting the arranged subviews all at once? I think it would increase readability by decreasing the scope needed to see which views are in the stackView and in which order.

theospears: Open to it, seems like a candidate for a follow up change.

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -167,26 +162,17 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
...
                     self.collectionView?.isHidden = true
                 case .UpdateSuggested:
-                    self.outofdateView.snp.remakeConstraints { (make) -> Void in
-                        make.left.equalTo(0)
-                        make.right.equalTo(0)
-                        make.top.equalTo(self.deadbeatView.snp.bottom)
-                        make.height.equalTo(42)
-                    }
+                    self.outofdateView.isHidden = false
                     self.outofdateLabel.isHidden = false
```
krugerk, line 171: must both be hidden simultaneously?

theospears: I think only the view. Removed the code to hide the label.

PR #545, opened 2024-12-06
gallery accounted for no longer displayed keyboard

krugerk:
## Summary
space could have been allocated for a keyboard that was not shown when returning to the gallery from a goal with the keyboard shown

follow-up to #536


theospears (approved):
This seems like a reasonable immediate term fix, thanks for putting it together.

PR #455, opened 2024-05-12
Remove the "delta text" UI section

theospears:
The app previously showed a section of numbers below the graph, e.g. (+1 +1 +2). These were from a time when "lanes" and "road" were part of the beeminder vocabulary and were pretty hard to understand, especially for do less goals.

We may wish to replace them with a table of what is due when, but for now just remove them.

Testing
Loaded the app and looked at a goal.

Issue #209, opened 2020-09-30
Due-by table


dreeves:
See how Beedroid does this now:

![image](https://user-images.githubusercontent.com/9928/94745710-1032a480-0330-11eb-9f12-1f6c6dbeb081.png)

The API returns `dueby` but it's currently undocumented. Discuss with @adamwolf.

(Alex Strick van Linschoten's feature request: Tell me how much I have to do to get into the green, etc.)

krugerk:
Beeminder for iOS does not currently display such a table but it does display the _deltatext_, colorized, for a goal (when viewing a single goal in the GoalVC and not the list of goals in the Gallery).

I see though [deltatext is marked as deprecated](https://api.beeminder.com/#attributes-2).

Example:
![IMG_F43FF9C41E06-1](https://user-images.githubusercontent.com/4656811/94747525-64b14080-037f-11eb-9744-f63ebcf7071f.jpeg)
where +2 and +6 appear under the graphs. 

The colors could be orange, blue, and green ,and correspond to each of the items in the deltatext.
First presented orange, second blue, third green.
today - orange - OK
tomorrow - blue - +2
in 2 days - green - +6


Does this mean the undocumented (and thus not officially supported) _dueby_ table should be used for the deprecated but documented (and thus officially supported) _deltatext_?

dreeves:
It's true that delta_text is deprecated and BeemiOS should stop using it. The dueby table (what's shown in the Statistics tab under the the graph on the website) is the future.

We definitely need to get dueby properly documented. We haven't yet because we expect to change how it works.

Side note: This may matter less with improvements to the safesum blurb that BeemiOS now shows at the top of the page ("Safe for 1 day or pay $5" in your screenshot). Namely, we should also prominently include the daily rate on that screen. That's mostly everything you need to know: (1) the exact amount due today if it's a beemergency, or (2) if you want to get N more safe days then just do N times the daily rate. If I'm committed to 1 vegetable/day and I'm in the orange and I want to get green then I should just eat 2 more vegetables. No need to quibble that technically I'm a quarter vegetable from being blue and can thus get in the green with 1.25 more vegetables. Just eat 2 vegetables and you know you're fine.

I mean, the dueby table is awesome if you want the full details laid out, but it's not information you need prominently on the goal screen or the dashboard.

strickvl:
[...]

Issue #36, opened 2020-04-27
Deltas under graph seem wrong


dreeves:
Placeholder, needs repro steps, etc, but this is from Grayson who's super clueful:

https://forum.beeminder.com/t/ios-app-graphs-show-oldest-4-out-most-recent-5-data-points/6631/6?u=dreev

dreeves:
More from Grayson on this:

https://forum.beeminder.com/t/ios-app-graphs-show-oldest-4-out-most-recent-5-data-points/6631/8?u=dreev

andrewpbrett:
So right now on grayson/junk-free:

```
> g = findg "grayson/junk-free"
> g.delta_text
=> "+1.18 +1.18 +1.18"
```

The deltas under the graph on the iOS app are based on what comes back from the API for `delta_text`. 

Could this actually be a beebrain/YBHP issue @dreeves @adamwolf ?

dreeves:
Ah, yes! So `delta_text` is a crude confusing version of the dueby table and maybe clients like BeemiOS should just parse `dueby` instead. (And better yet, philosophically, I think BeemiOS should just show safesum and make it easy to get to the full dueby table if you want to know the things delta_text was telling you.) But also maybe that's important for Beeminder to keep `delta_text` correct for backwards compatibility. Namely, the amounts you need to do get 1, 2, and 3 days of safety buffer. (It was only ever a vague approximation to that pre-YBHP. Lanes delenda est!)

Aaand, dammit, `dueby` isn't actually documented in the API. Maybe it's there though, just undocumented?

So, ok, the API needs to not break that. Gissued as https://github.com/beeminder/beeminder/issues/1350 -- stand by.

dreeves:
This should be done now!

andrewpbrett:
Thanks! I'm going to close this then; definitely agree that it would be better to just have the dueby table in the app. Probably a larger discussion though - it might violate PDP to remove it, especially if there's not a dueby table somewhere in the app to replace it.

PR #547, opened 2024-12-10
Restore the Deltas! GoalVC includes deltas again, now sourced from the dueby

### LINKED ISSUE #559: All goal deltas appear to be missing when opening up individual goal pages.
# Environment 

iOS 18.2
Beeminder iOS app version 6.7 (54)
iPhone 13 Pro

# Replicata

1. From the home screen of iOS [while logged into Beeminder iOS app to an account with at least one active goal] --  tap the icon to open the Beeminder iOS app --> Goals Page
2. Goals Page --> Goal
3. Arrive at [Actual] Resultata

# Expectata [Result]

Expectata taken from first App Store demonstration image - goal deltas are present.

image

# [Actual] Resultata

No goal deltas are present.

image

krugerk:
## Summary
Includes the due by table in the goal, displaying the deltas on the goal screen, where delta_text was previously displayed.
The user can quickly see, at the goal screen, how many of a thing to do to obtain a number of days of safety.

due by deltas

and now colored
due by deltas, colored
due by deltas, colored, on light

## Validation
built and ran the app in the simulator
viewed the goal screen of various goals

Relates to #209 
This shows the deltas on the goal screen. It does not display the entire table and it does not show the totals nor does it show the dates. It restores "amounts to do" previously provided by "delta text". (See #455)
Fixes #559


BeeKit/Model/Goal.swift
```diff
@@ -184,3 +201,113 @@ public class Goal: NSManagedObject {
...
+// Helper methods for conversion
+extension Goal {
+    func saveDueByTable(_ dictionary: [String: BeeminderDueByEntry]) {
+        do {
+            let data = try JSONEncoder().encode(dictionary)
+            if let jsonString = String(data: data, encoding: .utf8) {
+                self.dueByTableData = jsonString
+            }
+        } catch {
+            print("Error encoding dictionary: \(error)")
```
krugerk, line 214: logging instead of mere printing

BeeSwift/GoalViewController.swift
```diff
@@ -488,7 +501,15 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
     func updateInterfaceToMatchGoal() {
         self.datapointTableController.hhmmformat = goal.hhmmFormat
         self.datapointTableController.datapoints = goal.recentData.sorted(by: {$0.updatedAt < $1.updatedAt})
+        
+        self.dueByLabel.text = goal.dueBy
+            .sorted(using: SortDescriptor(\.key))
+            .compactMap { $0.value.formatted_delta_for_beedroid }
+            .map { $0 == "✔" ? "✓" : $0 }
```
krugerk, line 508: preferring this other checkmark, which shows up well on both light and dark

It is also the same used to indicate a goal has data for the day, today (when "todayta" is true, see https://github.com/beeminder/BeeSwift/blob/745cf9875ed6b0303c46d5d8e4c0a2d44796040d/BeeSwift/GoalCollectionViewCell.swift#L27)

dreeves:
I don't love the aesthetics of this. Maybe also in terms of user-experience, it's cluttery and overwhelming and opaque.

krugerk:
> I don't love the aesthetics of this. Maybe also in terms of user-experience, it's cluttery and overwhelming and opaque.

The expectation is that the goal screen contains an expandable section, maybe just a label "due by table" where upon tapping it, it expands to show the full table? It seems you may have suggested as much previously https://github.com/beeminder/BeeSwift/issues/36#issuecomment-627741771 .

BeeKit/Model/Goal.swift
```diff
@@ -62,6 +62,14 @@ public class Goal: NSManagedObject {
     @NSManaged public func addToRecentData(_ values: Set<DataPoint>)
     @objc(removeRecentData:)
     @NSManaged public func removeFromRecentData(_ values: Set<DataPoint>)
+    
+    @NSManaged public var dueByTableData: String?
+    
+    public var dueBy: [String: BeeminderDueByEntry] {
+        self.loadDueByTable() ?? [:]
+    }
```
theospears, line 70: Can we not use a [ValueTransformer](UICollectionViewDiffableDataSource) to avoid maintaining both properties?

krugerk: I think you meant to link to something else with the ValueTransformer link there. I wanted to avoid making a separate core data entity for the due by table. I can look into the value transformer.

theospears:
Sharing my higher level thoughts here
* We clearly need to provide this date (or at least a subset of it) to users in the app
* I'm not convinced this is the right UI
* But once the data model parts of this PR are in a good place, it seems reasonable that we merge it to be a starting point to iterate the UI on.

BeeKit/Model/Goal.swift
theospears: You're right, I was trying to link to https://developer.apple.com/documentation/coredata/handling_different_data_types_in_core_data#3582174

Although https://developer.apple.com/documentation/foundation/valuetransformer might have been a better link

krugerk:
> I'm not convinced this is the right UI

For all of the users of the app on iOS, as is, the deltas shown basically look like they did when the users knew them from delta_text. So any user on iOS should be automatically familiar with this interface.

The MR is here as a "request for comments/feedback/suggestions/...". May its spark encourage more and may it evolve from creation til merge.

BeeKit/Model/Goal.swift
krugerk: done

BeeSwift/GoalViewController.swift
```diff
@@ -515,7 +528,10 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
     func updateInterfaceToMatchGoal() {
         self.datapointTableController.hhmmformat = goal.hhmmFormat
         self.datapointTableController.datapoints = goal.recentData.sorted(by: {$0.updatedAt < $1.updatedAt})
-
+        
+        self.deltasLabel.isHidden = goal.dueByTable.entries.isEmpty
```
theospears, line 532: [Nit] This won't collapse the height - maybe an argument for StackView here as well. Doesn't seem like a big deal though.

BeeKit/Model/Goal.swift
```diff
@@ -62,6 +62,11 @@ public class Goal: NSManagedObject {
     @NSManaged public func addToRecentData(_ values: Set<DataPoint>)
     @objc(removeRecentData:)
     @NSManaged public func removeFromRecentData(_ values: Set<DataPoint>)
+    
+    @NSManaged public var dueByTable: DueByTable
```
theospears, line 66: I suggest calling this variable `dueBy` - I've been trying to exactly match api names (modulo casing and underscores) so I don't need to remember both.

BeeKit/Model/DueBy/DueByTable.swift
```diff
@@ -0,0 +1,38 @@
...
+        let entries = coder.decodeObject(forKey: Key.entries.rawValue) as? [String: BeeminderDueByEntry] ?? [:]
+        
+        self.init(entries: entries)
+    }
+    
+    convenience init(dueByJson: JSON?) {
+        var entries: [String : BeeminderDueByEntry] {
+            dueByJson?.dictionary?.compactMapValues(BeeminderDueByEntry.init)
+            ?? dueByJson?.dictionary?.mapValues(BeeminderDueByEntry.init)
+            ?? [:]
```
theospears, line 34: Could you explain what is happening here?

BeeKit/Model/DueBy/DueByTable.swift
```diff
@@ -0,0 +1,38 @@
...
+    
+    private enum Key: String {
+        case entries
+    }
+    
+    public init(entries: [String: BeeminderDueByEntry]) {
+        self.entries = entries
+    }
+    
+    public static var supportsSecureCoding: Bool { true }
```
theospears, line 18: From: https://developer.apple.com/documentation/foundation/nssecurecoding

> An object that does override [init(coder:)](https://developer.apple.com/documentation/mapkit/mkmapitem/init(coder:)) must decode any enclosed objects using the [decodeObjectOfClass:forKey:](https://developer.apple.com/documentation/foundation/nscoder/1442558-decodeobjectofclass) method. For example:

It appears we aren't doing this, so shouldn't be returning true here?

BeeKit/Model/DueBy/BeeminderDueByEntry.swift
```diff
@@ -0,0 +1,56 @@
...
+    public static var supportsSecureCoding: Bool { true }
+
+    private enum Key: String {
+        case total
+        case delta
+        case formattedTotalForBeedroid
+        case formattedDeltaForBeedroid
+    }
+    
+    public func encode(with coder: NSCoder) {
```
theospears, line 35: Do we need all this custom encode/decode logic? My understanding is Swift (4.0+) should be able to synthesize it.

theospears (changes requested):
Thanks for updating this to ValueTransformer. 
The most important bit of feedback here is: Do we actually need to have all the custom coding/decoding logic, or can the swift compiler generate it for us?
https://developer.apple.com/documentation/foundation/archives_and_serialization/encoding_and_decoding_custom_types#2904055

BeeKit/Model/DueBy/DueByTable.swift
krugerk: True. It seems I missed one entry.

BeeKit/Model/DueBy/BeeminderDueByEntry.swift
krugerk: Codable can take care of some of it. The rest is for NSCoding (NSSecureCoding). We can make one use the other, though this is not available out of the box, afaik. Will look into it

BeeSwift/GoalViewController.swift
```diff
@@ -656,3 +672,42 @@ private extension GoalViewController {
         return UIMenu(title: "bmndr.com/\(goal.owner.username)/\(goal.slug)", children: actions)
     }
 }
+
+
+private extension GoalViewController {
+    var dueByContainsSpecificAmounts: Bool {
```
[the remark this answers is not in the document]
krugerk: It is. The idea was just to show the deltas when at least one of them was a number (not just all checkmarks). I am thinking it might be better to just leave the checkmarks and thus this code remains unused and can be removed.

BeeSwift/AppDelegate.swift
```diff
@@ -31,6 +31,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             UIFont.beeminder.defaultFontPlain.withSize(20)]
         UIBarButtonItem.appearance().setTitleTextAttributes([NSAttributedString.Key.font : UIFont.beeminder.defaultFontPlain.withSize(18)], for: UIControl.State())
 
+        DueByTableValueTransformer.register()
```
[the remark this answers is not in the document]
krugerk: Yes, this makes sense - move it to the location where the rest of core data is initialized.

BeeKit/Model/DueBy/BeeminderDueByEntry.swift
theospears: If we are using NSSecureCoding, can we use `NSSecureUnarchiveFromDataTransformer` instead of a custom one?

krugerk: I do not think so. I think if custom, core data non-native, are being stored, then a custom valuetransformer is needed.

krugerk: > Do we need all this custom encode/decode logic?

This uses the pattern the rest of the app uses: SwiftyJSON and manually plucking out from the json.

> My understanding is Swift (4.0+) should be able to synthesize it.

NSCoding's encode: `public func encode(with coder: NSCoder)`
Codable's encode: `public func encode(to encoder: Encoder) throws`

theospears:
I put together an example of what I was thinking with regards to using Codable in #617.

krugerk:
> I put together an example of what I was thinking with regards to using Codable in #617.

OK. This merge request has been updated accordingly: removed the intermediary container that just stored the dictionary. Removed the NS(Secure)Coding of `BeeminderDueByEntry`. Used lowerCamelCase instead of snake_case.

BeeKit/Model/Goal.swift
```diff
@@ -62,6 +62,8 @@ public class Goal: NSManagedObject {
     @NSManaged public func addToRecentData(_ values: Set<DataPoint>)
     @objc(removeRecentData:)
     @NSManaged public func removeFromRecentData(_ values: Set<DataPoint>)
+    
+    @NSManaged public var dueByDaystamp: DueByDictionary
```
krugerk, line 66:
```suggestion
    @NSManaged public var dueByDay: DueByDictionary
```

```suggestion
    @NSManaged public var dueByDayMap: DueByDictionary
```

```suggestion
    @NSManaged public var dueByDate: DueByDictionary
```

BeeSwift/GoalViewController.swift
krugerk: I would hope the app's UI would all be written in `SwiftUI` before long. Shall the views first be migrated to `UIKit`'s `UIStackView`?

theospears (approved):

krugerk:
Another suggestion was showing

> something like

> 1 safe day
> [delta for the first non-safe day]

PR #554, opened 2024-12-16
badge count not reset upon logout

### LINKED ISSUE #553: badge not being set (reset) on log out
The badge shows the number of _beemergencies_ for the day. Upon signing out, the count should arguably be 0 since the app no longer refers to any particular account.

AppDelegate executes `updateBadgeCount` on both 'goals updated' and 'user signed out'. Now in the latter case, the call to obtain the goals returns `nil` and the badge count is not set.

https://github.com/beeminder/BeeSwift/blob/b55316207fb12dba832f70ad87a403f13d000ebf/BeeSwift/AppDelegate.swift#L136

Version: 6.7 (54)

krugerk:
## Summary
Badge count is not being set when a user logs out. 

## Validation
Ran the app in the simulator. Signed in with account that had beemergencies. Noted the badge count. Signed out. Noticed the badge count. 

## Issue
Fixes #553

### DIFF
--- a/BeeSwift/AppDelegate.swift
+++ b/BeeSwift/AppDelegate.swift
@@ -133,8 +133,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         assert(Thread.isMainThread, "updateBadgeCount must be run on the main thread")
 
         let context = ServiceLocator.persistentContainer.viewContext
-        guard let goals = ServiceLocator.goalManager.staleGoals(context: context) else { return }
-        let beemergencyCount = goals.count(where: { $0.safeBuf < 1})
+        
+        let beemergencyCount = ServiceLocator.goalManager
+            .staleGoals(context: context)?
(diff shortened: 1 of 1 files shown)


BeeSwift/AppDelegate.swift
```diff
@@ -133,8 +133,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         assert(Thread.isMainThread, "updateBadgeCount must be run on the main thread")
 
         let context = ServiceLocator.persistentContainer.viewContext
-        guard let goals = ServiceLocator.goalManager.staleGoals(context: context) else { return }
```
krugerk, line 136: Another approach would be to execute another code block on sign out, setting the badge count directly to 0 without first looking up goals in the db.

theospears: No strong opinions either way. But if doing it this way, let's add a comment explaining it is also intended to cover the logged out case.

Issue #88, opened 2020-07-09
Include toothbrushing


andrewpbrett:
Toothbrushing time is, apparently, a metric in Apple Health: https://developer.apple.com/documentation/healthkit/data_types. It's a duration/event, which has given us some issues in the past, but it would be nice to include it.

krugerk:
Would we be able to extract both of the following?
 - number of minutes brushed (per day)
 - number of teethbrushing sessions (per day)

Would the app grab the data then need to manage it itself in order to provide both of the above?

dreeves:
Got another request for this today. User apolyton has an Oral-B toothbrush that records minutes spent each day brushing teeth (recommended: 2 minutes twice a day so 4 minutes) and sends it to Apple Health.

PR #556, opened 2024-12-17
toothbrushing as seconds per day

krugerk:
## Summary
Users have requested beeminding toothbrushing and using Apple Health and the Beeminder iOS app to do so.
As is, this is an MVP of #88 in that it tracks toothbrushing as "total number of seconds" per day. Fancier would be for the integration to support
 
- [ ] Number of minutes brushed (per day)
- [ ] Number of teethbrushing sessions (per day)

directly.

## Validation
Added data to Apple Health in the Simulator
Ran the app in the simulator. Set up a goal, linked it with Toothbrushing, let it log datapoints.


krugerk:
Once code has run with this and then another build of the app is used where the app does not yet support the metric, there are errors and warnings about the missing metric as well as crashes in places like `RemoveHKMetricViewController` where `self.goal.humanizedAutodata` is being force unwrapped.

BeeKit/HeathKit/HealthKitMetric.swift
```diff
@@ -12,6 +12,7 @@ public enum HealthKitCategory : String, CaseIterable {
     case Heart = "Heart"
     case Mindfulness = "Mindfulness"
     case Nutrition = "Nutrition"
+    case SelfCare = "Self Care"
```
theospears, line 15: These categories map to the categorizations in the Health app. Toothbrushing is under "Other".

BeeKit/HeathKit/HealthKitConfig.swift
```diff
@@ -51,6 +51,10 @@ public enum HealthKitConfig {
             QuantityHealthKitMetric(humanText: "Vitamin K", databaseString: "dietaryVitaminK", category: .Nutrition, hkQuantityTypeIdentifier: .dietaryVitaminK),
             QuantityHealthKitMetric(humanText: "Water", databaseString: "water", category: .Nutrition, hkQuantityTypeIdentifier: .dietaryWater),
             
+            // Self care
+            ToothbrushingDailyMinutesHealthKitMetric.make(),
+            ToothbrushingDailySessionsHealthKitMetric.make(),
```
theospears, line 56: I don't want to have multiple top level metrics for the same underlying metric. There are lots of metrics where this could make sense, and the list risks getting far too long.
The way I think this should work:
1. There is one metric for toothbrushing.
2. On the preview/confirm page, there are additional settings to refine exactly how the metric is logged. These are stored using additional attributes on the autodata key on the goal. The preview updates as these settings are changed.
3. For toothbrushing in particular, this should be either "Aggregate Total Daily Minutes", or reporting each toothbrushing session as a separate data point with its number of minutes. People can count sessions via adjusting aggday.

Note this will require some tweaks to the sync code to allow multiple data points per day to be synced correctly. A while back the app started including requestid with these points to help facilitate this.

BeeKit/HeathKit/Toothbrushing/ToothbrushingDailyMinutesHealthKitMetric.swift
```diff
@@ -0,0 +1,36 @@
...
+        HKUnit.second()
+    }
+    
+    static func make() -> ToothbrushingDailyMinutesHealthKitMetric {
+        .init(humanText: "Teethbrushing (in seconds per day)",
+              databaseString: healthkitMetric,
+              category: HealthKitCategory.SelfCare)
+    }
+    
+    override func recentDataPoints(days: Int, deadline: Int, healthStore: HKHealthStore) async throws -> [any BeeDataPoint] {
```
theospears, line 27: Rather than this, can probably just implement `valueInAppropriateUnits`

BeeKit/HeathKit/Toothbrushing/ToothbrushingDailyMinutesHealthKitMetric.swift
```diff
@@ -0,0 +1,36 @@
...
+    
+    override func units(healthStore : HKHealthStore) async throws -> HKUnit {
+        HKUnit.second()
+    }
+    
+    static func make() -> ToothbrushingDailyMinutesHealthKitMetric {
+        .init(humanText: "Teethbrushing (in seconds per day)",
+              databaseString: healthkitMetric,
+              category: HealthKitCategory.SelfCare)
+    }
```
theospears, line 25: Let's remove this. I'd like all these constants to remain in `HealthKitConfig.swift` for consistency.

BeeSwift/Settings/RemoveHKMetricViewController.swift
```diff
@@ -44,7 +44,7 @@ class RemoveHKMetricViewController: UIViewController {
             attrString.append(NSMutableAttributedString(string: "\(self.goal.slug)\n",
                 attributes: [NSAttributedString.Key.font: UIFont.beeminder.defaultBoldFont]))
             
-            attrString.append(NSMutableAttributedString(string: "This goal obtains its data from Apple Health (\(self.goal.humanizedAutodata!)). You can disconnect the goal with the button below.",
+            attrString.append(NSMutableAttributedString(string: "This goal obtains its data from Apple Health (\(self.goal.humanizedAutodata ?? self.goal.healthKitMetric ?? "unknown metric")). You can disconnect the goal with the button below.",
```
theospears, line 47: Thanks for fixing this.

theospears (changes requested):

BeeKit/HeathKit/HealthKitConfig.swift
krugerk: Customizing `aggday` is possible with a custom goal. A typical do more (hustler) goal type will have its _how to aggregate the day_ set to sum.

BeeKit/HeathKit/Toothbrushing/ToothbrushingDailyMinutesHealthKitMetric.swift
krugerk: This was implemented for the purpose of specifying the healthkitMetric in the comment.

Issue #563, opened 2024-12-25
Make Beeminder dark mode splash screen consistent with other Dark Mode PR.


emeraldfleur:
Hello.

Not a bug, but I'll treat it like a "defect" for the issue writeup.

# Environment

iOS 18.2, iPhone 16 Pro Simulator
Dark Mode, Not Logged into Beeminder app

iOS 18.2, iPhone 13 Pro
Dark Mode, Logged into Beeminder app.

# Steps to Replicata

- Open Beeminder App
- **Arrive at Splash Screen**
- Arrive at Login Page

I've screenshotted the splash screen in dark mode in simulator, but I was able to confirm the above replicates on device as well in the second test environment listed above.
![Simulator Screenshot - iPhone 16 Pro - 2024-12-25 at 09 44 20](https://github.com/user-attachments/assets/5347658f-362c-4dc4-9cd7-585b4d024953)

I will work to submit a PR with an infinibee dark mode icon for comparison, that might look better on OLED iPhones where the contrast ratio is an asset.

Thank you.

emeraldfleur:
If I'm not mistake, this same image is also used as a graph placeholder, which looks better in context, one assumes.. will take before and after screenshots.

emeraldfleur:
Submitted PR #564 for this issue.

emeraldfleur:
Fixed by #564! :-)

PR #561, opened 2024-12-24
Changed Color of Beeminder Logo Text on Login Screen to White in Dark Mode

emeraldfleur:
## Summary

Changed color of Beeminder logo text on iOS app login screen to white in dark mode, by modifying relevant JSON to pull the dark mode .png file added in dark mode. Improves readability, and subjectively improves aesthetics. (Edit: Fixes #150)

Before:

(Screenshot taken on iOS 18.2, Beeminder iOS app version 6.7 (54), iPhone 13 Pro)

image
After:
image

Could not validate on hardware due to pending Apple Developer Account application, but literally just a flattened PNG image swap with same dimensions.

## Validation

Code compiles and app launches on iPadOS and iPhone 16 Pro simulators.

Could not validate on hardware due to pending Apple Developer Account application.

PR #564, opened 2024-12-25
Changed Splash Screen and Placeholder Graph Images to match Login Page Dark Mode PR

emeraldfleur:
## Summary
Changed splash screen and placeholder / thumbnail graph images to match recent Login Page Dark Mode PR #561, for consistency. Fixes new issue #563.

Before:

Simulator Screenshot - iPhone 16 Pro - 2024-12-25 at 09 44 20

After:

Simulator Screenshot - iPhone 16 Pro - 2024-12-25 at 10 23 07

Subjectively, this seems to be better to match the other pending dark mode PR, #561. This also changes the placeholder image while waiting for an internet connection to load a graph, but due to being dark, your eye kind of glosses over it. Couldn't get a screenshot of placeholder image while loading graph because it went by so fast. 

## Validation
App launches on device (iOS 18.2 iPhone 13 Pro)  and in simulator. (iOS 18.2 iPhone 16 Pro) There are no code changes, merely .png asset swaps with no changes to the identifying .json files, and the new PNGs are the same dimensions as the old ones.


theospears (approved):
LGTM!

theospears:
This change really makes me want dark-mode graphs. But not enough to perform atrocities like client-side image patching. We'll just have to wait for @dreeves to add support for this server side.

theospears:
Ok, it turns out to be surprisingly hard to get the app build to notice that the images have changed. In future, let's avoid changing .png files, and instead add new files with new names, and update the config to point at them.

krugerk:
> This change really makes me want dark-mode graphs. But not enough to perform atrocities like client-side image patching. We'll just have to wait for @dreeves to add support for this server side.

Others have called for having a table like listing and skipping the graphs altogether in the gallery. We could look into that, perhaps showing  something like what the ios 14 widgets MR presents.

emerald-pham:
> Ok, it turns out to be surprisingly hard to get the app build to notice that the images have changed. In future, let's avoid changing .png files, and instead add new files with new names, and update the config to point at them.
> 
> 
> 
> See http://arsenkin.com/launch_screen_image_cache.html for more details.

Will do. Yeah, got used to clearing my cache before a build, oops 😅

emeraldfleur:
I can add a new PR to change the file names so that it gets noticed on subsequent builds, if desired? Not a hard PR, and might count as an issue...

theospears:
> I can add a new PR to change the file names so that it gets noticed on subsequent builds, if desired? Not a hard PR, and might count as an issue...

I would accept such a PR, but I don't see it as very important. It will fix itself over time as people restart their devices.

PR #574, opened 2024-12-28
Deprecate storing user data in UserDefaults

theospears:
## Summary
Previously we were double-writing to CoreData and to UserDefaults in case we decided not to move ahead with CoreData. Now we have a public release that is fully migrated to CoreData we are committed to the migration, and don't need to maintain the UserDefaults paths.

This leaves in the code which migrates data from UserDefaults to CoreData, so it it still possible to seamlessly upgrade from previous versions, but removes incremental updates to UserDefaults, and deletes all values after the data has been migrated. This means downgrading from this version to <6.7 will require logging in again.

## Validation
* [x] Launch the app and verify you can reload goals, view timezone
* [x] Verify it is still possible to seamlessly upgrade from code before the CoreData migration

### DIFF
--- a/BeeKit/Managers/CurrentUserManager.swift
+++ b/BeeKit/Managers/CurrentUserManager.swift
@@ -48,11 +48,13 @@ public actor CurrentUserManager {
 
     init(requestManager: RequestManager, container: BeeminderPersistentContainer) {
         self.requestManager = requestManager
-        modelContainer = container
+        self.modelContainer = container
         let context = container.newBackgroundContext()
         context.name = "CurrentUserManager"
-        modelExecutor = .init(context: context)
+        self.modelExecutor = .init(context: context)
+
         migrateValuesToCoreData()
+        cleanUpUserDefaults()
     }
 
     // If there is an existing session based on UserDefaults, create a new User object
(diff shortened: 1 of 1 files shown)


theospears:
* **#574** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/574?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/574?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

BeeKit/Managers/CurrentUserManager.swift
```diff
@@ -82,6 +83,14 @@ public actor CurrentUserManager {
         try! context.save()
     }
 
+    private nonisolated func cleanUpUserDefaults() {
+        for key in CurrentUserManager.allKeys {
+            userDefaults.removeObject(forKey: key)
```
krugerk, line 88: though this removes not just elements of user but also the access token; not sure if that was intended

theospears: Not strictly intended but I think fine - the access token should be stored in the keychain now.

PR #576, opened 2024-12-28
Resolve cached images not updating on build by updating two Content.json files to point to four renamed .png file names.

emeraldfleur:
## Summary

#564 had an issue where swapping only the .png files in place did not replace the images stored in cache.

This PR appends the word, "new", to the end of four .png filenames, and updates two Content.json files to point to these renamed .png files to ensure they get noticed on build...

## Validation

Assets confirmed loading in local copy of simulator (iPhone 16 Pro 18.2) on local copy of Repo, did **not** test for transition from cached assets to non-cached assets on build.

image


krugerk:
🤔 as an alternative how about a gh action that checks merge requests for adjusted images, commenting about the cache or suggesting using new names?

theospears (approved):
Let's see if this helps!

emeraldfleur:
> 🤔 as an alternative how about a gh action that checks merge requests for adjusted images, commenting about the cache or suggesting using new names? 

I was going to open an issue like normal, but figured it'd be a one off. May create an issue based on that idea though for future tracking...

theospears:
> I was going to open an issue like normal, but figured it'd be a one off. May create an issue based on that idea though for future tracking...

As a general philosophy, I'd suggest not being over-zealous in creating issues. My reasoning:
* If today we think this is important, we should just do it
* If we don't think it's important today, and nothing changes, we'll probably never decide to do the work to address the issue
* If something does happen to make us think this is important then we can take action in response to that, and thus don't need the issue.

emeraldfleur:
I'll probably open issues on this repo only in the case where a PR for a non-existing issue might take more than a day to complete and I can't get the PR done on the same day to prevent someone else from working on the same issue accidentally. (Whenever there's not even enough material for a draft PR...) Thank you for the note Theo!

PR #579, opened 2024-12-29
list does not match website when sorting by pledge

### LINKED ISSUE #578: app's list when sorted by pledge is unpredictable and unlike the website's list
The app allows sorting goals by name (from a to z), by urgency (from most urgent to least), by deadline (from sooner to later), and by recent data (from most recently updated to least recently updated).
When selecting to sort the gallery by pledge, the goals are listed top to bottom from the ones with the highest pledge to the ones with the least. This list however, does not seem stable nor does it correspond with the one the website shows. The website appears to sort by decreasing pledge amounts and uses urgency (most to least urgent) for tiebreakers (when goals share the same pledge amount).

I suspect the same would have been true for other sort criteria that may have had ties. Names would be unique (no two goals can have the same name for a given user). Deadline and the timestamp of when last touched (recent data) could have also bee the same for multiple goals.

app: 6.7 (54)
website: v2024.12.28-23dbd

krugerk:
## Summary
When sorting by pledge, goals were sorted grouped by pledge amounts. The sort order within seemed undefined.

Also used the opportunity to clean the code somewhat using `sorted(using:)` where multiple comparators may be provided.

## Validation
Ran app in simulator.
Compared before and after to list when sorting on the website.

Fixes #578

### DIFF
--- a/BeeSwift/Gallery/GalleryViewController.swift
+++ b/BeeSwift/Gallery/GalleryViewController.swift
@@ -366,7 +366,9 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
                     return goal1.lastTouch > goal2.lastTouch
                 }
                 else if selectedGoalSort == Constants.pledgeGoalSortString {
-                    return goal1.pledge > goal2.pledge
+                    return goal1.pledge == goal2.pledge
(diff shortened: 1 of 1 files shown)


theospears (approved):
Mich nicer!

Issue #360, opened 2023-01-05
App claims not to know user's timezone


dreeves:
A user reports that their timezone is listed in BeemiOS settings as "Unknown" for some unknown reason:

<img width="320px" src="https://user-images.githubusercontent.com/9928/210670387-b65ad0a2-3b35-423b-af61-b49497e1e1ef.png"/>

## Desiderata

- [ ] Discuss how impactful this is, does it replicate for us?
- [ ] TBD

## Cognata

* https://secure.helpscout.net/conversation/2112981500/141009?folderId=2972058
* #282
* #31
* https://github.com/beeminder/BeeSwift/blob/master/BeeSwift/SettingsViewController.swift#L117
* https://forum.beeminder.com/t/ios-app-ignoring-time-changes-to-deadlines/9221/2

_**Verbata**: timezone, time zone,_

krugerk:
Unknown was shown when the data meant to be saved could not be retrieved from UserDefaults.

#594 fetches from the User endpoint upon opening the Settings screen and when the user performs a Pull-To-Refresh there.

How about marking this ticket as 'could not reproduce' or similar?

krugerk:
Could be marked as fixed with the release of [6.8](https://github.com/beeminder/BeeSwift/releases/tag/builds%2Fiosbeta_ci%2F79) to the [App Store](https://apps.apple.com/de/app/beeminder/id551869729)

PR #594, opened 2024-12-31
settings shows user's timezone, provide ptr to fetch user

### LINKED ISSUE #507: App doesn't notice when you switch Beeminder timezone
```[tasklist]
### Desiderata
- [ ] Nail down replicata
```

User moved from Chicago to New York, changed their timezone in Beeminder, but then BeemiOS was confused for months until they finally logged out and logged back into the app.

## Cognata

* #361 
* #360 
* #282 
* https://secure.helpscout.net/conversation/2663990321/192046?folderId=2972058

_**Verbata**: timezone hell,_

krugerk:
## Summary
Settings shows the user's timezone but only a few places in the app were fetching the user: fetching default notification settings (entering the default notification settings screen) and upon signing in.
This MR allows the user to trigger a fetch of the User endpoint via a Pull-To-Refresh on the Settings screen.

settings - ptr

## Validation
used the app in the simulator
change the timezone at the website
tapped around in the app again both entering settings and pull to refresh from settings

Fixes #507 and helps out with #360

### DIFF
--- a/BeeSwift/Settings/SettingsViewController.swift
+++ b/BeeSwift/Settings/SettingsViewController.swift
@@ -92,7 +92,7 @@ class SettingsViewController: UIViewController {
     }
     
     @objc private func fetchUser() {
-        Task { [weak self] @MainActor in
+        Task { @MainActor [weak self] in
             self?.tableView.isUserInteractionEnabled = false
             try? await ServiceLocator.currentUserManager.refreshUser()
             
--- a/BeeSwift/Settings/SettingsViewController.swift
+++ b/BeeSwift/Settings/SettingsViewController.swift
@@ -92,7 +92,7 @@ class SettingsViewController: UIViewController {
     }
     
     @objc private func fetchUser() {
-        Task { [weak self] in
+        Task { [weak self] @MainActor in
             self?.tableView.isUserInteractionEnabled = false
             try? await ServiceLocator.currentUserManager.refreshUser()
             
--- a/BeeSwift/Settings/SettingsViewController.swift
+++ b/BeeSwift/Settings/SettingsViewController.swift
@@ -34,10 +34,13 @@ class SettingsViewController: UIViewController {
         
         self.tableView.delegate = self
         self.tableView.dataSource = self
-        self.tableView.isScrollEnabled = false
         self.tableView.tableFooterView = UIView()
         self.tableView.register(SettingsTableViewCell.self, forCellReuseIdentifier: self.cellReuseIdentifier)
-
+        self.tableView.refreshControl = {
+            let refreshControl = UIRefreshControl()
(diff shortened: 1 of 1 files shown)


BeeSwift/Settings/SettingsViewController.swift
```diff
@@ -87,6 +90,19 @@ class SettingsViewController: UIViewController {
     func showLogsTapped() {
         self.navigationController?.pushViewController(LogsViewController(), animated: true)
     }
+    
+    @objc private func fetchUser() {
+        Task { [weak self] in
```
krugerk, line 95:
```suggestion
        Task { [weak self] @MainActor in
```

theospears:
I have a PR in the works which will update User every time goals are updated (as a side effect of moving to use diff_since and incremental goal update). If that change is merged, would you still want to do this?

krugerk:
> I have a PR in the works which will update User every time goals are updated (as a side effect of moving to use diff_since and incremental goal update). If that change is merged, would you still want to do this?

It is a very basic implementation as is. Of course if user is being fetched more often anyway, then one like this at the settings screen is perhaps too basic. 
I think it is fine like this for now and maybe yours with using diff_since makes it moot.

BeeSwift/Settings/SettingsViewController.swift
theospears: I assume I should wait for you to either commit this or resolve this conversation thread before merging.

theospears, line 95: There's some interesting nuance to lifetimes here.

It might be tempting to do this:
```
          guard let self = self else { return }
```
But it's not quite right, as the view might be dismissed while waiting for the refreshUser call to complete, and in that case we don't want to hold on to `self`. 

Maybe worth a comment making this clear?

BeeSwift/Settings/SettingsViewController.swift
```diff
@@ -87,6 +90,19 @@ class SettingsViewController: UIViewController {
     func showLogsTapped() {
         self.navigationController?.pushViewController(LogsViewController(), animated: true)
     }
+    
+    @objc private func fetchUser() {
+        Task { [weak self] @MainActor in
```
krugerk, line 95:
```suggestion
        Task { @MainActor [weak self] in
```

BeeSwift/Settings/SettingsViewController.swift
krugerk: as in,

```swift
@objc private func fetchUser() {
        Task { [weak self] in
            self?.tableView.isUserInteractionEnabled = false

            try? await ServiceLocator.currentUserManager.refreshUser()

            guard let self else { return }
            self.tableView.reloadData()
            self.tableView.layoutIfNeeded()
            self.tableView.refreshControl?.endRefreshing()
            
            self.tableView.isUserInteractionEnabled = true
        }
    }

```

PR #591, opened 2024-12-31
Fix GoalViewController memory leak

theospears:
## Summary
We were retaining references to GoalViewController meaning it was not correctly cleaned up when leaving the view. There were two causes of this.
1. A reference cycle between `GoalViewController` and `DatapointTableController` (fixed via weak ref)
2. A Timer maintaining a reference to the view (fixed via invalidating the timer on hide).

Fixes #590

## Validation
Ran the app in the simulator. Navigated to the goal view and back several times. Ran the XCode memory graph debugger.
Before this patch: Observed many GoalViewController still in memory
After this patch: No copies of the class in memory

Logged out and logged into the app. Observed there was no longer a crash reading goal objects.

Issue #590, opened 2024-12-30
GoalViewController crashes when logging out and then logging in again


theospears:
From https://github.com/beeminder/BeeSwift/pull/537#discussion_r1899483977

unfortunately that has the side effect of the app crashing when accessing goal.lastModifiedLocal at GoalViewController.updateLastUpdatedLabel upon signing into the app.

    GoalManager.performPostGoalUpdateBookkeeping posts notification goalsUpdated
    GoalViewController.onGoalsUpdatedNotification executes
    calling its updateInterfaceToMatchGoal
    GoalViewController.updateInterfaceToMatchGoal calls, among others, updateLastUpdatedLabel
    GoalViewController.updateLastUpdateLabel now accesses self.goal.lastModifiedLocal, which at the time, is still part of an empty NSManagedObject Goal:

<Goal: 0x600002148aa0> (entity: Goal; id: 0xblahblahblah9494 <x-coredata://blahblah-F9F9-4FB3-BD1B-blahblah/Goal/p385>; data: {
    alertStart = nil;
    autodata = nil;
    data = "<relationship fault: 0x600000383740 'data'>";
    deadline = nil;
    graphUrl = nil;
    healthKitMetric = nil;
    hhmmFormat = nil;
    id = nil;
    initDay = nil;
    lastModifiedLocal = nil;
    lastTouch = nil;
    leadTime = nil;
    limSum = nil;
    owner = nil;
    pledge = nil;
    queued = nil;
    recentData = "<relationship fault: 0x6000003829a0 'recentData'>";
    safeBuf = nil;
    safeSum = nil;
    slug = nil;
    thumbUrl = nil;
    title = nil;
    todayta = nil;
    urgencyKey = nil;
    useDefaults = nil;
    won = nil;
    yAxis = nil;
})

theospears:
The XCode memory debugger suggests the view is being kept around by references from an NSTimer (presumably for `updateLastUpdatedLabel`) and NotificationCenter (presumably for `onGoalsUpdatedNotification`)

<img width="1013" alt="image" src="https://github.com/user-attachments/assets/b8215250-d995-45b9-b6d7-1f7f84a29fcd" />

krugerk:
I still see it. 

Steps to reproduce:
logged in, with goals
from the gallery, open a goal
from the goal, return to the gallery
this triggers automatically a fetch goals and updates the db
open another goal (in particular, before the gallery has been reloaded with goals after their recent fetch)
[...]

PR #596, opened 2025-01-01
goalvc was crashing when refreshing its ui for an unavailable goal

krugerk:
## Summary
Despite the changes in #591, a crash was still observed. 

## Validation
added logging to validate vc being deallocated when expected
in simulator, navigated gallery to goal and back and forth again

#590

### DIFF
--- a/BeeSwift/GoalViewController.swift
+++ b/BeeSwift/GoalViewController.swift
@@ -26,7 +26,7 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
     
     private let timeElapsedView = FreshnessIndicatorView()
     fileprivate var goalImageView = GoalImageView(isThumbnail: false)
-    fileprivate var datapointTableController = DatapointTableViewController()
+    fileprivate var datapointTableController: DatapointTableViewController?
     fileprivate var dateTextField = UITextField()
     fileprivate var valueTextField = UITextField()
     fileprivate var commentTextField = UITextField()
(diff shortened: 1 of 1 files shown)


BeeSwift/GoalViewController.swift
```diff
@@ -495,9 +501,9 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
         updateInterfaceToMatchGoal()
     }
 
-    func updateInterfaceToMatchGoal() {
-        self.datapointTableController.hhmmformat = goal.hhmmFormat
-        self.datapointTableController.datapoints = goal.recentData.sorted(by: {$0.updatedAt < $1.updatedAt})
+    @MainActor func updateInterfaceToMatchGoal() {
+        self.datapointTableController?.hhmmformat = goal.hhmmFormat
+        self.datapointTableController?.datapoints = goal.recentData.sorted(by: {$0.updatedAt < $1.updatedAt})
```
krugerk, line 506:
```suggestion
        self.datapointTableController?.datapoints = goal.recentData.sorted(using: SortDescriptor(\.updatedAt))
```

BeeSwift/GoalViewController.swift
```diff
@@ -312,6 +315,9 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
         super.viewDidDisappear(animated)
         lastUpdatedTimer?.invalidate()
         lastUpdatedTimer = nil
+        datapointTableController?.delegate = nil
+        datapointTableController = nil
+        NotificationCenter.default.removeObserver(self)
```
krugerk, line 320:
```suggestion
```

docs point out that modern iOS versions take care of this automatically

theospears:
Are you able to reliably repro the crash? If so, what does the memory debugger show at the time? I'm very open to the idea this patch is correct, but would like to understand why.

PR #462, opened 2024-07-08
Store goal information in CoreData

theospears:
Introduce a new Goal type in CoreData. Update the goal fetching code to also write all fetched goal data to this store, along side the in memory store. At present app functionality is still reading from the in memory store rather than core data.

Testing:
Make sure the app still launches, can show the goal list, and view individual goals

PR #597, opened 2025-01-02
error downloading goal graph failing url ?procTime=

krugerk:
## Summary
The NSManagedObject Goal can itself be non-nil while containing non-nil but empty graph and thumbnail url strings. Although this state was possible at code level before the transition to storing goals in core data, it had not occurred then and has been encountered regularly since.

error seen in log:
```
Error downloading goal graph: AlamofireImage.AFIError.alamofireError(Alamofire.AFError.sessionTaskFailed(error: Error Domain=NSURLErrorDomain Code=-1002 "unsupported URL" UserInfo={NSLocalizedDescription=unsupported URL, NSErrorFailingURLStringKey=?proctime=, NSErrorFailingURLKey=?proctime=,
```

## Validation
watched the log for output in Xcode
signed into an account with goals
signed out

App Version: any since using Core Data (#462) including 6.8 (58)


krugerk:
At the code level, this can still happen. However, with #607 I have not seen it.

PR #599, opened 2025-01-02
meta datapoints being deleted or overwritten

### LINKED ISSUE #598: HealthKit integration removes meta datapoint RESTARTED ON THE
Upon restarting a goal via the website, a datapoint is added to the goal to indicate as much. This has the syntax _RESTARTED ON THE {ordinal day number}_

These are currently being overwritten by the HealthKit integration, which attempts to be itself the authority over an associated goal's datapoints.

Notes:
The app does currently handle datapoints containing #RESTARTED:
https://github.com/beeminder/BeeSwift/blob/741432b254b68d05e93daa4b9f78522d94131f9d/BeeKit/Model/DataPoint.swift#L100

Maybe this is something that could be _fixed_ server side, since currently the server does not use the # symbol.

##### Versions
App: 6.8 (58)
Website: v2025.01.01-7a966

krugerk:
## Summary
Meta datapoints refer to those beeminder itself creates to indicate things like goal has been restarted, when a goal has been created, when a derailment has occurred and so on. The app differentiates these so as to not delete them when syncing data to goals from HealthKit. Currently the app does not consider origin.

## Validation
Had a goal with a manually created datapoint with one of the known strings.
Same goal also had a similar datapoint generated from beeminder.
Goal was integrated with HealthKit.

Would fix #598


BeeKit/Model/DataPoint.swift
```diff
@@ -103,6 +107,7 @@ extension DataPoint {
     /// DataPoints are used to track certain events, like automatic pessimistic values, goal restarts, derailments, etc. These should sometimes
     /// be treated differently, e.g. not deleted as part of syncing with HealthKit
     public var isMeta: Bool {
-        DataPoint.metaPointHashtags.contains { comment.contains($0) }
+        guard origin == "beeminder" else { return false }
```
krugerk, line 110: _ex nihilo_ also
> "ex nihilo" is the datapoint added by Beeminder when a goal for the first time

BeeKit/Model/DataPoint.swift
```diff
@@ -103,6 +107,7 @@ extension DataPoint {
...
+        return DataPoint.metaPointHashtags.contains { comment.contains($0) }
```
krugerk, line 111: I am thinking checking origin is enough and the comment content check can be skipped. The website is origin beeminderweb or similar.

theospears:
@dreeves how would you feel about making a small server side tweak to add `dummy (boolean)` and possibly `initial (boolean)` to the datapoint resource in the API, rather than the iOS app reproducing this logic?

krugerk:
> @dreeves how would you feel about making a small server side tweak to add `dummy (boolean)` and possibly `initial (boolean)` to the datapoint resource in the API, rather than the iOS app reproducing this logic?


How come?

bsoule:
> @dreeves how would you feel about making a small server side tweak to add dummy (boolean) and possibly initial (boolean) to the datapoint resource in the API, rather than the iOS app reproducing this logic?

adding fields to the resource is cheap and not harmful to any of the code we're maintaining, i don't think.  i guess `initial` is just `origin == nihilo`? what do you mean by/about `dummy` however?

theospears:
I hear the datapoint now has is_dummy and is_initial, so let's use those and get rid of the isMeta helper entirely.
Probably the behavior should be slightly different in different places, e.g. apple health sync should ignore both, and the default input value should ignore is_dummy but include is_initial (e.g. for lose weight goals)

krugerk:
> I hear the datapoint now has is_dummy and is_initial, so let's use those and get rid of the isMeta helper entirely. Probably the behavior should be slightly different in different places, e.g. apple health sync should ignore both, and the default input value should ignore is_dummy but include is_initial (e.g. for lose weight goals)

seems you took care of this in #651

PR #600, opened 2025-01-02
clean code around NSNotification.Name

krugerk:
## Summary
Some places defined strings that were then meant to be converted to NSNotification.Name. This MR removes that step so the `NSNotification.Name`s can be used directly.

## Validation
Could use as before, notifications such as signing out and goals updated were still observed.


krugerk:
At the same time, I think the app should move away from using these, especially those deal with goals being updated.

theospears (approved):

Issue #215, opened 2020-10-12
App should respect time format settings


krugerk:
When displaying "xx units due by", the app does not respect the user's preferred time formatting.

 - Configure the iOS device where the app is installed to prefer a 24 hour mode.

Notice the app continues displaying times with am and pm: "xx units due by 12am".

 
The app should respect the user's preference as indicated by the settings on the device. The following order may present a sensible 

 - use the device's setting "24 hour yes/no" set in date and time preferences
 - use the device's timezone setting
 - use the user's timezone setting, set in the user's beeminder account

![IMG_5816](https://user-images.githubusercontent.com/4656811/95718314-8c9d7f80-0c6e-11eb-9cf8-2a5b33768f67.jpg)

krugerk:
See also #233

PR #601, opened 2025-01-02
show when healthkit linked goal was last synced with healthkit

### LINKED ISSUE #193: app should indicate in the app when it was last provided data from apple health
Probably it should do this per metric / goal. As is (Version 5.7), one could infer this via the datapoints seen in the account at the website or in the app.
It would be beneficial though if the app itself kept track of when it last received data from Apple Health and display that to the user.

An idea would be to remove the two sync buttons in the goal view, replacing them with one sync button (sync last two weeks9 and another area displaying when the last update was received.

krugerk:
## Summary
The user can see the most recent datapoints and can infer from that when the last sync took place. This PR displays the date directly so the user need not infer when that might have been.

Simulator Screenshot - iPhone 12 Pro

## Validation
ran app in simulator

## Tickets
Fixes #193

### DIFF
--- a/BeeSwift/GoalViewController.swift
+++ b/BeeSwift/GoalViewController.swift
@@ -38,6 +38,7 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
     fileprivate var countdownLabel = BSLabel()
     fileprivate var scrollView = UIScrollView()
     fileprivate var submitButton = BSButton()
+    private let pullToRefreshView = PullToRefreshView()
     fileprivate let headerWidth = Double(1.0/3.0)
     fileprivate let viewGoalActivityType = "com.beeminder.viewGoal"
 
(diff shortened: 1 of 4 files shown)


BeeSwift/GoalViewController.swift
```diff
@@ -361,6 +357,21 @@ class GoalViewController: UIViewController,  UIScrollViewDelegate, DatapointTabl
         self.countdownLabel.textColor = self.goal.countdownColor
         self.countdownLabel.text = self.goal.capitalSafesum()
     }
+    
+    private func refreshPullDown() {
+        let dateFormatter = ISO8601DateFormatter()
+        dateFormatter.formatOptions = [.withFullDate, .withFullTime]
```
krugerk, line 363: #215

dreeves:
Proposal: strip that down to nothing but HH:MM -- no date, no seconds, no am/pm, no timezone, nothing. I would've thought that too extreme but it's what Stripe does on their mobile app and I realized it works because you figure it out by experimentation.

krugerk:
> Proposal: strip that down to nothing but HH:MM -- no date, no seconds, no am/pm, no timezone, nothing. I would've thought that too extreme but it's what Stripe does on their mobile app and I realized it works because you figure it out by experimentation.

updated to show date too (in user's preferred format) when last sync was not in the same day and to show the time in the user's preferred format

krugerk:
Still marked as draft because as is a goalVC will not update as the goal does to show the more recent last synced time. It just shows it as it was when the gaolVC was opened.

Issue #588, opened 2024-12-30
gallery seems slow to load and displays conflicting data


krugerk:
The "last updated" label on the gallery often shows a more recent relative time than the goals listed underneath it in that same moment. Then, some while later, while the label still shows the same "last updated", the set of goals are refreshed, showing their more up-to-date states.

From the sign in screen: 
fill out username and password, tap the Sign In button
Time 0 

The Sign In screen is still shown but now with a disabled Sign In Button and with an activity indicator.

Time 1.1s
the sign in screen starts sliding out of view to reveal the gallery

Gallery's last updated is shown and it shows 57 seconds ago

The gallery appears otherwise empty (no goals shown and no message about not having any goals). The activity indicator is shown.

Time 1.6s
Gallery now reports last updated now and shows the message for not having any goals. The activity indicator is still shown. This account has goals.

Time 4.5s
Gallery no longer shows the activity indicator, the goals are now shown all at once, the last updated still (or again?) shows now

The gallery could have a better (initial) loading state. Let this be a separate topic.

There appears to be some lag between when the gallery shows "last updated now" and when the contents of the collection view show the goals in their "now" state. This discrepancy is something new that came about with the integration of goals in core data (6.7 (51)).

Once the user has signed in, the app would have received the list of goal names for non-archived, non-deleted goals. The app triggers a `signedInNotification` which the Gallery receives. It triggers its fetchGoals which in turn has the GoalManager fetch from the goals endpoint, persisting the data locally in core data.
It seems to me that at least initially all of the last updated + list of goals + activity indicator should all update in the same moment.

theospears:
I've also noticed this. I've been unable to obviously see the cause from looking through the code.

krugerk:
It seems particularly noticeable when bringing the app into the foreground after a while in the background when the signed in account has HealthKit linked goals. The gallery shows goal cells and it can be scrolled but the scrolling is laggy or choppy.

krugerk:
In the simulator this issue seemed no longer present with changes from #607. Once a build with the changes is available for hardware (a build in TestFlight), the situation will be checked again. Perhaps those changes fixed this issue altogether.

krugerk:
A user on the forum reported what appears to be the same behavior: https://forum.beeminder.com/t/ios-ux-comments/12243/3.

krugerk:
Most of what this ticket describes has been addressed, is no longer reproducible, with the recent version available in TestFlight: 6.8 (62).

[...]

PR #607, opened 2025-01-05
Gallery's list is now fed by nsfetchedresultscontroller

krugerk:
## Summary
The initial loading (displaying) of goals on the gallery was slow. The app was still relying on passing and receiving nsnotification/notifications.
With the data in Core Data, the NSFetchedResultsController can be used.

Migrated galleryVC to use the fetched results controller.
Updated the sorting implementation accordingly, directly applying the sort type to the fetchRequest itself.
Updated the filtering as well to also apply directly to the fetchRequest.

UI wise this looks the same, just faster/more responsive.

## Validation
Ran the app in the simulator.
Changed sorting.
Changed filters.
Signed in.
Signed out.
Navigated from Gallery to Goal and back.
Sent app into background. Brought back into foreground.
Force quit app, relaunched app.

## Tickets
helps with #588

### DIFF
--- a/BeeSwift/Gallery/GalleryViewController.swift
+++ b/BeeSwift/Gallery/GalleryViewController.swift
@@ -67,7 +67,7 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
         self.healthStoreManager = healthStoreManager
         
         let fetchRequest = Goal.fetchRequest() as! NSFetchRequest<Goal>
-        fetchRequest.sortDescriptors = [NSSortDescriptor(keyPath: \Goal.urgencyKey, ascending: true)]
+        fetchRequest.sortDescriptors = Self.preferredSort
         fetchedResultsController = .init(fetchRequest: fetchRequest,
                                          managedObjectContext: viewContext,
                                          sectionNameKeyPath: nil, cacheName: nil)
@@ -350,7 +350,7 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
         self.didUpdateGoals()
     }
     
-    private var preferredSort: [NSSortDescriptor] {
+    static private var preferredSort: [NSSortDescriptor] {
         let selectedGoalSort = UserDefaults.standard.value(forKey: Constants.selectedGoalSortKey) as? String ?? Constants.urgencyGoalSortString
         
         switch selectedGoalSort {
@@ -383,7 +383,7 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
             self.fetchedResultsController.fetchRequest.predicate = nil
         }
         
-        self.fetchedResultsController.fetchRequest.sortDescriptors = preferredSort
+        self.fetchedResultsController.fetchRequest.sortDescriptors = Self.preferredSort
         try? self.fetchedResultsController.performFetch()
     }
     
--- a/BeeSwift/Gallery/GalleryViewController.swift
+++ b/BeeSwift/Gallery/GalleryViewController.swift
@@ -40,9 +40,6 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
     let searchBar = UISearchBar()
     var lastUpdated: Date?
     
-    var goals : [Goal] = []
-    var filteredGoals : [Goal] = []
-    
     private enum Section: CaseIterable {
         case main
     }
@@ -52,8 +49,11 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
     public enum NotificationName {
         public static let openGoal = Notification.Name(rawValue: "com.beeminder.openGoal")
     }
-
-    init(currentUserManager: CurrentUserManager, 
+    
+    private let fetchedResultsController: NSFetchedResultsController<Goal>!
+    private var fetchRequest: NSFetchRequest<Goal>?
+    
+    init(currentUserManager: CurrentUserManager,
          viewContext: NSManagedObjectContext,
          versionManager: VersionManager,
          goalManager: GoalManager,
@@ -63,7 +63,16 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
         self.versionManager = versionManager
         self.goalManager = goalManager
         self.healthStoreManager = healthStoreManager
+        
+        let fetchRequest = Goal.fetchRequest() as! NSFetchRequest<Goal>
+        fetchRequest.sortDescriptors = [NSSortDescriptor(keyPath: \Goal.urgencyKey, ascending: true)]
+        fetchedResultsController = .init(fetchRequest: fetchRequest,
+                                         managedObjectContext: viewContext,
+                                         sectionNameKeyPath: nil, cacheName: nil)
+        self.fetchRequest = fetchRequest
+        
         super.init(nibName: nil, bundle: nil)
+        fetchedResultsController.delegate = self
     }
     
     required init?(coder: NSCoder) {
@@ -208,6 +217,9 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
                 logger.error("Error checking for current version: \(error)")
             }
         }
+        
+        try? fetchedResultsController.performFetch()
+        
     }
     
     override func viewWillAppear(_ animated: Bool) {
@@ -235,10 +247,8 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
     }
     
     @objc func handleGoalsFetchedNotification() {
-        Task {
-            self.lastUpdated = await goalManager.goalsFetchedAt
-            self.updateGoals()
-        }
+        logger.debug("\(#function)")
+        logger.debug("doing nothing because GalleryVC should be informed over Core Data about updates to goals of interest")
     }
     
     @objc func settingsButtonPressed() {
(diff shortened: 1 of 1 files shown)


BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -17,10 +17,85 @@ import CoreData
...
+            
+            collectionViewOps.append(BlockOperation(block: { [weak self] in
+                self?.collectionView?.deleteItems(at: [indexPath])
+            }))
+        case .update:
+            guard let indexPath = indexPath else { break }
+            
+            collectionViewOps.append(BlockOperation(block: { [weak self] in
+                // self?.collectionView?.deleteItems(at: [indexPath])
+                // self?.collectionView?.insertItems(at: [indexPath])
```
krugerk, line 54:
```suggestion
```

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -55,7 +137,19 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
         self.versionManager = versionManager
         self.goalManager = goalManager
         self.healthStoreManager = healthStoreManager
+        
+        let fetchRequest: NSFetchRequest<Goal> = NSFetchRequest<Goal>(entityName: Goal.entity().name!)
+        fetchRequest.sortDescriptors = [NSSortDescriptor(keyPath: \Goal.urgencyKey, ascending: true)]
```
krugerk, line 142: use the preferred sort from the get go


```suggestion
        let fetchRequest: NSFetchRequest<Goal> = NSFetchRequest<Goal>(entityName: Goal.entity().name!)
        fetchRequest.sortDescriptors = preferredSort
```

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -17,10 +17,83 @@ import CoreData
 
 import BeeKit
 
+extension GalleryViewController: NSFetchedResultsControllerDelegate {
+    func controllerDidChangeContent(_ controller: NSFetchedResultsController<NSFetchRequestResult>) {
```
theospears, line 21: What are your thoughts on using `UICollectionViewDiffableDataSource` instead of managing the diff process by hand?

krugerk: I have implemented this as its separate pull request. See #614.

krugerk: updated based on changes in #614
maybe that one be merged first and I shall adjust this one accordingly

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -504,3 +524,17 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
         self.fetchGoals()
     }
 }
+
+extension GalleryViewController: NSFetchedResultsControllerDelegate {
```
theospears, line 528: I would be open to the idea of having all the various delegate interfaces implemented by `GalleryViewController` added as part of extensions, but right now _every other interface_ is included in the initial class definition, and I was confused why `NSFetchedResultsControllerDelegate` wasn't.

[Do] Let's be consistent for now and include this interface in the main class definition. We can split it out in a later PR with all the others.

```swift
class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayout, UICollectionViewDelegate, UISearchBarDelegate, SFSafariViewControllerDelegate, NSFetchedResultsControllerDelegate
```

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -63,7 +63,16 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
         self.versionManager = versionManager
         self.goalManager = goalManager
         self.healthStoreManager = healthStoreManager
+        
+        let fetchRequest = Goal.fetchRequest() as! NSFetchRequest<Goal>
+        fetchRequest.sortDescriptors = [NSSortDescriptor(keyPath: \Goal.urgencyKey, ascending: true)]
```
theospears, line 68: For self: is this sometimes the wrong sort?

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -208,6 +217,9 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
                 logger.error("Error checking for current version: \(error)")
             }
         }
+        
+        try? fetchedResultsController.performFetch()
```
theospears, line 221: Nit: Is it worth logging if this fails?

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -504,3 +524,17 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
         self.fetchGoals()
     }
 }
+
+extension GalleryViewController: NSFetchedResultsControllerDelegate {
+    func controllerDidChangeContent(_ controller: NSFetchedResultsController<NSFetchRequestResult>) {
+        self.applySnapshot()
+    }
+    
+    func controller(_ controller: NSFetchedResultsController<NSFetchRequestResult>,
```
theospears, line 533: Do we need this?

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -504,3 +524,17 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
         self.fetchGoals()
     }
 }
+
+extension GalleryViewController: NSFetchedResultsControllerDelegate {
+    func controllerDidChangeContent(_ controller: NSFetchedResultsController<NSFetchRequestResult>) {
+        self.applySnapshot()
```
theospears, line 530: [Consider]
https://developer.apple.com/videos/play/wwdc2019/230/?time=890 suggests that we should be able to get a snapshot from the NSFetchedResultsController and apply it directly to the `DiffableDataSource`.

image

https://stackoverflow.com/a/58475482 suggests that this can work, although you need to use `UITableViewDiffableDataSource<String,NSManagedObjectID>`

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -340,46 +348,53 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
...
+            self.fetchedResultsController.fetchRequest.predicate = nil
         }
+        
+        self.fetchedResultsController.fetchRequest.sortDescriptors = preferredSort
+        try? self.fetchedResultsController.performFetch()
+        
+        self.applySnapshot()
     }
     
     private func applySnapshot() {
```
theospears, line 394: See below about getting snapshots directly from the FetchedResultsController.

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -340,46 +348,53 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
...
-            self.filteredGoals = self.goals.filter { (goal) -> Bool in
-                return goal.slug.lowercased().contains(searchText.lowercased())
-            }
+            self.fetchedResultsController.fetchRequest.predicate = nil
         }
+        
+        self.fetchedResultsController.fetchRequest.sortDescriptors = preferredSort
+        try? self.fetchedResultsController.performFetch()
+        
+        self.applySnapshot()
```
theospears, line 391: Do we need to `applySnapshot()` here, or will the delegate callback method handle that?

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -271,8 +281,6 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
     }
     
     @objc func handleSignOut() {
-        self.goals = []
-        self.filteredGoals = []
         self.applySnapshot()
```
theospears, line 284: This should no longer be needed, right?

theospears (changes requested):
Things are going to be so much cleaner without all the manual bookkeeping around goals/core data changing.

Added some notes. I think the most interesting one is around having NSFetchedResultsController create snapshots directly. I think we should at least try that, although it's possible it will end up being worse.

BeeSwift/Gallery/GalleryViewController.swift
krugerk: It would have defaulted to urgencyKey then pulled from the user's preference right away. Yes it could have been wrong for a moment initially.
Updated to just use the preference also here.

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -482,18 +483,20 @@ class GalleryViewController: UIViewController, UICollectionViewDelegateFlowLayou
     }
     
     private func configureDataSource() {
-        let cellRegistration = UICollectionView.CellRegistration<GoalCollectionViewCell, Goal> { cell, indexPath, goal in
+        let cellRegistration = UICollectionView.CellRegistration<GoalCollectionViewCell, NSManagedObjectID> { [weak self] cell, indexPath, goalObjectId in
+            let goal = self?.fetchedResultsController.fetchedObjects?.first(where: { $0.objectID == goalObjectId })
```
theospears, line 487: Why this over `fetchedResultsController.object(at: indexPath)` ?

krugerk, line 487:
```suggestion
            let goal = self?.fetchedResultsController.object(at: indexPath)
```

krugerk: no good reason. changed

theospears (approved):

BeeSwift/Gallery/GalleryViewController.swift
krugerk: Made a MR for these: #615

PR #616, opened 2025-01-15
place gallery sort on gallery

krugerk:
## Summary
As is, the sorting preference of the gallery is located in a screen accessible from the settings screen. Urgency might be the sort order most commonly used (hypothesis) with recent data also making sense as a way to check which goals have recently been updated. Changing from to the other requires four taps and takes place across three screens.
This Merge Request proposes offering changing the sorting preference of the gallery while still at the gallery. It reduces the taps from four to two.

button sort option
list of sort options

## Validation
Ran app in simulator. 
Sorted the gallery.


theospears (approved):
Looks reasonable to me. I see this is marked as a draft, do you think it is ready to merge, or does it need more work?

krugerk:
> Looks reasonable to me. I see this is marked as a draft, do you think it is ready to merge, or does it need more work?

I marked it as draft because as is a goalVC will not update as the goal does to show the more recent last synced time. It just shows it as it was when the gaolVC was opened.

PR #617, opened 2025-01-20
Minimal DueBy table using Codable and Json Encoding/Decoding

theospears:
## Summary
This is an example of an approach to storing the due by table with minimal manual code to handle encoding and decoding.

In particular, compared to #547
* It uses `Codable` to handle codegen for all serialization/deserializtion
* Removes the `DueByTable` type in favor of a direct dictionary, removing the need to reference via `entries`.

## Validation
Loaded the app and confirmed I could view the deltas on the goal page.


krugerk:
This one has become obsolete now that #547 has been merged.

PR #618, opened 2025-01-20
gallery not updating list based on sort

krugerk:
## Summary
Gallery was not updating how its list was sorted after the preference had been adjusted. Also, the gallery's last updated was meant to display when the goals were last fetched.

### 
The gallery was not monitoring for changes to the sorting preference. It needed a trigger to sort the list anew. This was removed in d181dba6da9a35b7b. 'Update goals' can stay, it is the fetch goals that was to be reduced. Furthermore, since gallery's last updated is meant to indicate how fresh the data is (when they were last fetched), gallery's setting this to now where it did does not necessarily coincide with fetching the goals anew. Instead, the goal manager already provides this nugget.

## Validation
Ran app in simulator. 
gallery -> settings -> sort pref, picked another, and back to gallery from settings
Noticed sort had taken effect.
Also noticed last updated seems much more up-to-date and less out of sync, as reported in #588.

### DIFF
--- a/BeeSwift/Gallery/GalleryViewController.swift
+++ b/BeeSwift/Gallery/GalleryViewController.swift
@@ -398,8 +398,12 @@ class GalleryViewController: UIViewController {
         self.collectionView.refreshControl?.endRefreshing()
         MBProgressHUD.hide(for: self.view, animated: true)
         self.updateDeadbeatVisibility()
-        self.lastUpdated = Date()
-        self.updateLastUpdatedLabel()
+        
+        Task {
+            self.lastUpdated = await goalManager.goalsFetchedAt
+            self.updateLastUpdatedLabel()
+        }
+        
         if self.filteredGoals.isEmpty {
             self.noGoalsLabel.isHidden = false
             self.collectionContainer.isHidden = true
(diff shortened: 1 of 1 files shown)


BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -275,10 +276,11 @@ class GalleryViewController: UIViewController {
     }
     
     override func viewDidAppear(_ animated: Bool) {
-        if !currentUserManager.signedIn(context: viewContext) {
+        guard currentUserManager.signedIn(context: viewContext) else {
             let signInVC = SignInViewController(currentUserManager: currentUserManager)
             signInVC.modalPresentationStyle = .fullScreen
             self.present(signInVC, animated: true, completion: nil)
+            return
```
theospears, line 283: Let's remove this part of the change - It's not relevant to the PR, and it's a kinda surprising use of guard.

theospears (changes requested):
Thanks for noticing this and fixing it. Let's remove the unrelated change but otherwise looks good. I've also poked the tests as I think the failure was spurious.

krugerk:
> Thanks for noticing this and fixing it. Let's remove the unrelated change but otherwise looks good. I've also poked the tests as I think the failure was spurious.

Yes it was. And my permissions on my account did not allow for my account to retrigger the action directly. Eventually a new post would happen to the branch (rebase off of recent master) and that would trigger the tests anew.

PR #625, opened 2025-01-26
Disable animation for goal thumbnails

theospears:
## Summary
Animation of goal thumbnails in the gallery often had some unfortunate affects:
* Because gallery cells are re-used, on returning to the gallery from a goal the gallery entry for a goal would show the graph _for a different goal_ before fading to the right one
* Refreshing the gallery would cause images to flicker to slightly darker and back

It may be possible to more carefully choose when to animate to address these, but for now avoid the issue by not animating thumbnails.

## Validation
Loaded the app and navigated to a goal and back. Observed the gallery immediately showed the correct graphs.

### DIFF
--- a/BeeSwift/Components/GoalImageView.swift
+++ b/BeeSwift/Components/GoalImageView.swift
@@ -77,8 +77,13 @@ class GoalImageView : UIView {
 
     @MainActor
     private func showGraphImage(image: UIImage) {
+        // Animating the thumbnail view interacts badly with cell re-use in the gallery
+        // e.g. it would cause us to show the image from a different goal before animating
+        // to the corrent one.
+        let duration = isThumbnail ? 0 : 0.4
+
         UIView.transition(with: imageView,
-                          duration: 0.8,
+                          duration: duration,
                           options: .transitionCrossDissolve,
                           animations: { [weak self] in
             self?.imageView.image = image


theospears:
* **#625** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/625?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/625?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

BeeSwift/Components/GoalImageView.swift
```diff
@@ -77,8 +77,13 @@ class GoalImageView : UIView {
 
     @MainActor
     private func showGraphImage(image: UIImage) {
+        // Animating the thumbnail view interacts badly with cell re-use in the gallery
```
krugerk, line 80: I would guess there is something wrong with the cell or the GoalImageView itself.

GoalCollectionViewCell has its [prepareForReuse](https://github.com/beeminder/BeeSwift/blob/638b04a8191709897938c33847a3a7f958541d79/BeeSwift/GoalCollectionViewCell.swift#L78) method which is called when the system is about to reuse a cell. The method should finish with the cell being in the state that it can be reused again. As is, it sets the `thumbnailImageView`'s `goal` to `nil`.

The expectation is probably that doing so would set the thumbnailImageView to [set the graph placeholder as the image](https://github.com/beeminder/BeeSwift/blob/638b04a8191709897938c33847a3a7f958541d79/BeeSwift/Components/GoalImageView.swift#L72) or at the very least removing the previously associated graph-thumbnail.

So what is going on in the lifecycle of GoalImageView? When is `clearGoalGraph` being called?

PR #632, opened 2025-02-21
Search goals also in title

### LINKED ISSUE #631: Search also in the goal description
As requested on the [forum](https://forum.beeminder.com/t/ios-ux-comments/12243/2), enhance the current filter/search by looking not only at goal names but also goal descriptions.

krugerk:
## Summary
In #631 it was requested that the search works should be extended to search also in goals' descriptions (here `title`s).

https://forum.beeminder.com/t/ios-ux-comments/12243

## Validation
Ran app in simulator. Searched goals for things appearing in titles and not in goal names. Searched goals for things appearing in names and not titles. Noticed (expected) results both times.

Fixes #631


BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -98,7 +98,7 @@ class GalleryViewController: UIViewController {
         let searchBar = UISearchBar()
         searchBar.accessibilityIdentifier = "searchBar"
         searchBar.delegate = self
-        searchBar.placeholder = "Filter goals by slug"
+        searchBar.placeholder = "Search goals"
```
krugerk, line 101: > Nit: on the Goals page, search is nice and snappy but the “Filter goals by slug” copy is really a back-of-house term, would recommend just “Search goals” instead.

 https://forum.beeminder.com/t/ios-ux-comments/12243

theospears (approved):

PR #635, opened 2025-02-24
Use incremental goal update

theospears:
## Summary
Previously the app would always fetch and update all goals on refresh. This was inefficient both in transfer time and amount of work the server needed to do, leading to slow updates for users with many goals. Now instead we ask the server only for updated goals, leading to a smaller diff.

## Validation
* Launched the app with a test account, loaded a goal, deleted it, confirmed it was gone
* Added unit tests to cover goal creation and deletion

### DIFF
--- a/BeeKit/Managers/GoalManager.swift
+++ b/BeeKit/Managers/GoalManager.swift
@@ -57,11 +57,6 @@ public actor GoalManager {
 
     /// Fetch and return the latest set of goals from the server
     public func refreshGoals() async throws {
-        guard let username = await currentUserManager.username else {
-            try await currentUserManager.signOut()
-            return
-        }
-
         guard let user = self.currentUserManager.user(context: modelContext) else { return }
         let goalsUnknown = user.goals.count == 0 || user.updatedAt.timeIntervalSince1970 == 0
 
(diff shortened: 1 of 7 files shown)


theospears:
* **#635** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/635?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/635?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

BeeKit/Managers/GoalManager.swift
```diff
@@ -57,17 +57,60 @@ public actor GoalManager {
...
+
+        let deleteMissingGoals: Bool
+        let deletedGoals: JSON
+
+        if (goalsUnknown) {
+            logger.notice("Goals unknown, doing full fetch")
+            // We must fetch the user object first, and then fetch goals afterwards, to guarantee User.updated_at is
+            // a safe timestamp for future fetches without losing data
+            userResponse = JSON(try await requestManager.get(url: "api/v1/users/{username}.json")!)
+            goalResponse = JSON(try await requestManager.get(url: "api/v1/users/{username}/goals.json")!)
```
krugerk, line 74: `goalResponse` here contains a list of goal objects as provided by the api.

BeeKit/Managers/GoalManager.swift
```diff
@@ -57,17 +57,60 @@ public actor GoalManager {
...
+
+            deleteMissingGoals = true
+            deletedGoals = JSON(arrayLiteral: [])
+        } else {
+            // TODO: Use old data to find deleted goals
+            logger.notice("Doing incremental update since \(user.updatedAt, privacy: .public)")
+            userResponse = JSON(try await requestManager.get(url: "api/v1/users/{username}.json", parameters: ["diff_since": user.updatedAt.timeIntervalSince1970 + 1])!)
+            goalResponse = userResponse["goals"]
```
krugerk, line 82: `goalResponse` here contains a list of strings, goal names, that have been modified since `diff_since`.

BeeKit/Managers/GoalManager.swift
```diff
@@ -95,33 +138,24 @@ public actor GoalManager {
             return
```
krugerk, line 138: When `updateGoalsFromJson` is called with the list of strings (goal names), this function exits early in this guard.

Where are the calls
`GET /users/u/goals/g.json`
for each of the goals corresponding to the goal names listed in the goalResponse when fetched via the user endpoint?

BeeKit/Managers/GoalManager.swift
```diff
@@ -57,17 +57,60 @@ public actor GoalManager {
...
 
-        let responseObject = try await requestManager.get(url: "api/v1/users/\(username)/goals.json")!
-        let response = JSON(responseObject)
-        self.updateGoalsFromJson(response) // TODO: Return failure info
+        // The user may have logged out during the network operation. If so we have nothing to do
+        modelContext.refreshAllObjects()
+        guard let user = modelContext.object(with: user.objectID) as? User else { return }
 
-        self.goalsFetchedAt = Date()
+        user.updateToMatch(json: userResponse)
```
krugerk, line 92: GoalManager now knows how to fetch user and set user in db. What is CurrentUserManager's role?

krugerk (commented):
When are the goals fetched when performing the incremental fetch?

BeeKit/Managers/GoalManager.swift
theospears: Sounds like
1) There's an opportunity for docs to be clearer here
2) It would help to add some unit tests illustrating the shape of goals in this case.

BeeKit/Managers/GoalManager.swift
theospears: I'm thinking GoalManager should turn into a general SyncManager which is responsible for all writes to CoreData. This will also avoid needing all the data refreshing logic. CurrentUserManager will just handle login/logout.

PR #636, opened 2025-02-24
Update freshness based on user last updated

theospears:
## Summary
We were fetching the last updated time for goals from a variable only stored in memory. This meant when launching the app it would show as 2024 years ago. Now we always update user every time we update goals, instead fetch the version we keep persisted on the user object.

We also replace all usages of refreshUser with refreshGoals so the two are always in sync.

## Validation
Hard quit the app. Launched and observed an appropriate value from a few seconds ago.

### DIFF
--- a/BeeKit/Managers/CurrentUserManager.swift
+++ b/BeeKit/Managers/CurrentUserManager.swift
@@ -108,21 +108,6 @@ public actor CurrentUserManager {
         return user(context: modelContext)?.username
     }
 
-    public func refreshUser() async throws {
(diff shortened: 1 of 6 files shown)


theospears:
* **#636** <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/636?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/beeminder/BeeSwift/636?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #646, opened 2025-04-09
alerts not shown when app in foreground

### LINKED ISSUE #460: Notifications don't come through for some users
## Desiderata

- [ ] Gather notes from discussion about device tokens

---

User ro_87 has for a long time now had problems with notifications coming through. I think she usually says she gets the first one in the day but no others.

Log excerpt from her phone:

```
[09/06/2024, 23:36] [AppDelegate] application:didFinishLaunchingWithOptions
[09/06/2024, 23:36] [AppDelegate] applicationDidBecomeActive
[09/06/2024, 23:36] [HealthStoreManager] Updating all goals with recent day for last 7 days
[09/06/2024, 23:36] [AppDelegate] application:didRegisterForRemoteNotificationsWithDeviceToken
[09/06/2024, 23:36] [AppDelegate] application:didReceiveRemoteNotification
```

image

## Cognata

* #50 
* #51 
* #96 

_**Verbata**: flaky notifications, reminders, zenos,_

krugerk:
## Summary
When the device receives a remote notification to display an alert, either 
a) the app is not in the foreground and the alert is shown in the Notification Center or
b) the app is in the foreground and thus far no alert was shown at all - neither in the app nor in iOS's Notification Center.

## Validation
Dropped a particularly formatted .apns json file onto the simulator.

Perhaps this fixes #460.

### DIFF
--- a/BeeSwift/AppDelegate.swift
+++ b/BeeSwift/AppDelegate.swift
@@ -17,7 +17,7 @@ import AlamofireNetworkActivityIndicator
 import BeeKit
 
 @UIApplicationMain
-class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterDelegate {
+class AppDelegate: UIResponder, UIApplicationDelegate {
     let logger = Logger(subsystem: "com.beeminder.beeminder", category: "AppDelegate")
     let backgroundUpdates = BackgroundUpdates()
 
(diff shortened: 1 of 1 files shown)


BeeSwift/AppDelegate.swift
```diff
@@ -41,6 +41,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         
         NetworkActivityIndicatorManager.shared.isEnabled = true
         
+        UNUserNotificationCenter.current().delegate = self
```
krugerk, line 44: > If your app is in the foreground when a notification arrives, the shared user notification center calls this method to deliver the notification directly to your app. If you implement this method, you can take whatever actions are necessary to process the notification and update your app. When you finish, call the completionHandler block and specify how you want the system to alert the user, if at all.
> If your delegate does not implement this method, the system behaves as if you had passed the **UNNotificationPresentationOptionNone** option to the completionHandler block. If you do not provide a delegate at all for the UNUserNotificationCenter object, the system uses the notification’s original options to alert the user.

- [userNotificationCenter(_:willPresent:withCompletionHandler:)](https://developer.apple.com/documentation/usernotifications/unusernotificationcenterdelegate/usernotificationcenter(_:willpresent:withcompletionhandler:))



https://github.com/beeminder/BeeSwift/blob/8384332f50bd9bdf26417637c86b8a6345ae9992/BeeSwift/AppDelegate.swift#L166-L168 has been implemented but was not being called since the app was not registered as the UNUserNotificationCenter's delegate.

theospears (approved):

PR #504, opened 2024-10-24
letting RelativeFormatter work its magic

krugerk:
also replaces the low contrast red on light grey with black on red

### before

red on gray

### after

black on red

PR #659, opened 2025-08-03
Make freshness indicator a bit fuzzy

theospears:
## Summary
The freshness indicator only updates from time to time, so showing a precise number of seconds is a bit misleading. It also leads to thrash when we update data a bunch of times in a row, each showing a slightly different value. Instead show an approximation of how long since last update.

## Validation
Loaded app on device, watched time slowly increase

### DIFF
--- a/BeeSwift/Components/FreshnessIndicatorView.swift
+++ b/BeeSwift/Components/FreshnessIndicatorView.swift
@@ -12,11 +12,8 @@ class FreshnessIndicatorView: UIView {
         return label
     }()
     
-    private let formatter: RelativeDateTimeFormatter = {
-        let formatter = RelativeDateTimeFormatter()
-        formatter.dateTimeStyle = .named
-        return formatter
-    }()
+    private var updateTimer: Timer?
+    private var lastUpdateDate: Date?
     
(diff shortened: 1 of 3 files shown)


krugerk:
If you recall, we had something similar before #504. You are right that the 'last updated ... ago' is not being updated itself often (every second) and thus itself can show stale data (updated 11 seconds when that was true when the view was created but not any longer true while the user sits at the screen). 

Maybe instead of introducing the date formatting logic removed with #504, we could take other approaches: one could be to show the fetched date: current as of <then showing the date and time>.

krugerk:
> Another option would be:

> 
> 2. Teach the app to poll and update itself every 5 minutes while open (so it mostly avoids the bar ever showing)
> 
> 



I had a branch (or even maybe a PR) that performed something like [zeno](https://blog.beeminder.com/zeno) [polling](https://blog.beeminder.com/glossary#z) for each goal: timed pull halfway between last pull/now and the goal's deadline. Coupled with other automatic polling (app enters foreground, app is informed of changed data thanks to push) and the user manually pulling-to-refresh, ... the app could show both when last fetched and when the next pull is scheduled.

PR #661, opened 2025-08-04
Add new App Intents

theospears:
## Summary
This deprecates the original AddData intent and adds new AddDataPoint and OpenGoal intents, backed by a GoalEntity

## Validation
* [x] Created and ran a shortcut which added data to a goal
* [x] Created and ran a shortcut which opened the app to a specific goal
* [x] Created and ran a shortcut which just opened the app

### DIFF
--- a/BeeSwift/Gallery/GalleryViewController.swift
+++ b/BeeSwift/Gallery/GalleryViewController.swift
@@ -23,6 +23,7 @@ class GalleryViewController: UIViewController {
     
     public enum NotificationName {
         public static let openGoal = Notification.Name(rawValue: "com.beeminder.openGoal")
+        public static let navigateToGallery = Notification.Name(rawValue: "com.beeminder.navigateToGallery")
     }
     
     // Dependencies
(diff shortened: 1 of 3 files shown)


BeeSwift/Intents/GoalEntityQuery.swift
```diff
@@ -0,0 +1,65 @@
...
+        
+        guard let currentUser = ServiceLocator.currentUserManager.user(context: context) else {
+            return []
+        }
+        
+        return try await context.perform {
+            let request = NSFetchRequest<Goal>(entityName: "Goal")
+            request.predicate = NSPredicate(format: "owner == %@", currentUser)
+            request.sortDescriptors = [NSSortDescriptor(key: "urgencyKey", ascending: true)]
+            request.fetchLimit = 20
```
krugerk, line 38: The drop down then only shows up to 20 entries and for one to have the shortcut open a goal not in the drop one, one would probably set the goalname inside a Text variable and reference that variable in the open action.

image

theospears: Claude worked it out. Fixed in https://github.com/beeminder/BeeSwift/pull/669

PR #671, opened 2025-09-10
Apply swift-format to the code base

theospears:
## Summary
Experimental PR to see what it looks like and whether we like it.


theospears:
Thoughts so far

### Positive
* Consistent import ordering
* Consistent trailing commas on lists
* More consistent empty lines

### Neutral
* Changed default indent
* Some changes to access modifiers

### Mixed
* Lots more line wrapping

BeeKit/Daystamp.swift
```diff
@@ -6,170 +6,178 @@ import Foundation
...
+  ///   - date: a calendar date
+  ///   - deadline: a ``Goal/deadline``
+  public init(fromDate date: Date, deadline: Int) {
+    let secondsAfterMidnight =
+      Daystamp.calendar.component(.hour, from: date) * 60 * 60
+      + Daystamp.calendar.component(.minute, from: date) * 60
+      + Daystamp.calendar.component(.second, from: date)
+
+    let dayOffsetFromDeadline =
+      if deadline < 0 {
```
theospears, line 57: Why did this get wrapped?

BeeKit/HeathKit/HealthKitConfig.swift
```diff
@@ -10,53 +10,107 @@ import Foundation
...
-            TimeAsleepHealthKitMetric(humanText: "Time asleep", databaseString: "timeAsleep", category: .Sleep),
-
-            // Other
-            QuantityHealthKitMetric(humanText: "Time in Daylight", databaseString: "timeInDaylight", category: .Other, hkQuantityTypeIdentifier: .timeInDaylight),
-        ]
-    }
+  public static var metrics: [HealthKitMetric] {
+    [
+      // Activity
+      QuantityHealthKitMetric(
```
theospears, line 16: This no longer reads as a table. OTOH the lines were _very_ long.

BeeSwift/Gallery/ChooseGoalSortViewController.swift
```diff
@@ -7,56 +6,56 @@
...
+    return Constants.goalSortOptions.count
+  }
+
+  func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+    return 1
+  }
+
+  func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+    guard let cell = tableView.dequeueReusableCell(withIdentifier: self.cellReuseIdentifier) as? SettingsTableViewCell
+    else { return UITableViewCell() }
```
theospears, line 48: The lack of any indentation on the follower line here is a bit surprising

BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -6,520 +6,552 @@
...
-            collectionView.dequeueReusableSupplementaryView(ofKind: kind, withReuseIdentifier: "footer", for: indexPath)
-        }
-        
-        self.collectionView.dataSource = dataSource
+
+    self.dataSource = .init(
+      collectionView: collectionView,
+      cellProvider: { collectionView, indexPath, goalObjectId in
+        collectionView.dequeueConfiguredReusableCell(
+          using: cellRegistration,
```
theospears, line 441: More wrapping here is better

BeeSwift/MainCoordinator.swift
```diff
@@ -1,240 +1,255 @@
...
+    let controller = EditGoalNotificationsViewController(
+      goal: goal,
+      currentUserManager: currentUserManager,
+      requestManager: requestManager,
+      goalManager: goalManager,
+      viewContext: viewContext)
+
+    navigationController.pushViewController(
+      controller,
+      animated: true)
```
theospears, line 174: Does this really need to wrap at all?

BeeSwift/SceneDelegate.swift
```diff
@@ -1,87 +1,92 @@
...
+
+    coordinator?.start()
+  }
+
+  func scene(_ scene: UIScene, continue userActivity: NSUserActivity) {
+    logger.info("\(#function)")
+
+    var userInfoOfGoalFromSpotlight: [AnyHashable: Any]? {
+      guard userActivity.activityType == CSSearchableItemActionType else { return nil }
+      guard let goalIdentifier = userActivity.userInfo?[CSSearchableItemActivityIdentifier] as? String else {
```
theospears, line 45: And here we _dont_ wrap the else, which is annoyingly inconsistent with the above case where we did.

BeeKit/Daystamp.swift
krugerk: Would you rather it stay as it was, have the block line up with the initial `if`, or something else?

```swift
 let dayOffsetFromDeadline = if deadline < 0 {
                                // This is an early deadline. If the time is after the deadline we need to instead consider it the next day
                                if secondsAfterMidnight >= Daystamp.secondsInDay + deadline {
                                    1
                                } else {
                                    0
                                }
                              } else {
                                  // This is a late deadline. If the time is before the deadline, we should consider it the previous day
                                  if secondsAfterMidnight < deadline {
                                      -1
                                  } else {
                                      0
                                  }
                              }
```

Ìt could also be rewritten to stick all of the block in its own space, something like
```swift
var dayOffsetFromDeadline: Int {
    if deadline < 0 {
        // This is an early deadline. If the time is after the deadline we need to instead consider it the next day
        if secondsAfterMidnight >= Daystamp.secondsInDay + deadline {
            1
        } else {
            0
        }
    } else {
        // This is a late deadline. If the time is before the deadline, we should consider it the previous day
        if secondsAfterMidnight < deadline {
            -1
        } else {
            0
        }
    }
}
```

BeeKit/HeathKit/HealthKitConfig.swift
krugerk: Maybe formatting these such that every parameter is its own line would be a way to go.

At the same time, it is not code that needs to be read or touched often.

BeeSwift/MainCoordinator.swift
krugerk: I thought perhaps this was affected by [linebreakbeforeeachargument](https://github.com/swiftlang/swift-format/blob/main/Documentation/Configuration.md#linebreakbeforeeachargument) being set to `false` in combination with the line being longer than 120 characters if placed horizontally but the line is 73 chars or so.
Nonetheless, I find it easier to read and more consistent with the rest of the file that such calls are split over multiple lines.

BeeSwift/SceneDelegate.swift
krugerk: It appears it can be forced by setting [linebreakbeforecontrolflowkeywords](https://github.com/swiftlang/swift-format/blob/main/Documentation/Configuration.md#linebreakbeforecontrolflowkeywords) to `true`.

krugerk:
> ### Neutral
> * Changed default indent

[indentation](https://github.com/swiftlang/swift-format/blob/main/Documentation/Configuration.md#indentation) defaults to 2 though could be configured to be 4 again.

krugerk (commented):
If _agents_ are writing the code, perhaps formatting is of less concern. However, when humans take a look at it, it sure would be nice to have consistent formatting at least within files and better yet across areas/targets.

PR #678, opened 2025-10-20
simulator in test is iphone17pro on ios26

krugerk:
## Summary
Following the iOS release (iOS 26), bump the simulator used in tests.

still using [Runner macOS-15](https://github.com/actions/runner-images/blob/main/images/macos/macos-15-Readme.md) and not yet [macOS-26-arm64](https://github.com/actions/runner-images/blob/main/images/macos/macos-26-arm64-Readme.md).


theospears:
`macos-15` is an arm image - see https://github.com/actions/runner-images/tree/main

PR #684, opened 2025-10-28
stabilizing fastlane-tests on ci

krugerk:
## Summary
The fastlane-tests GitHub workflow uses the macOS 15. Since #678 the workflow has failed often, timing out while apparently waiting for the simulator to start. May the combination macOS 26 + iOS 26 simulator run better.

## Validation
checking results in GitHub actions


krugerk:
> Based on the failing test status, I assume this remains a work in progress.

The culprit remains unclear to me.

theospears:
Did you learn anything here in terms of what does/does not mitigate the issue?

theospears:
Going to default to closing this out as it's not clear we reached a conclusion, but feel free to reopen.

PR #689, opened 2025-11-29
Add dietary fiber HealthKit metric

chipmanaged:
Adds dietary fiber as a new Apple Health nutrition metric (following the existing pattern for other nutrition metrics).

Testing: Tested in simulator. Might be worth testing on a physical device first. (I'm not yet able to do that (sorry!) but will do so in the future, after I have the requestSigningKey and all that.)

### DIFF
--- a/BeeKit/HeathKit/HealthKitConfig.swift
+++ b/BeeKit/HeathKit/HealthKitConfig.swift
@@ -120,6 +120,12 @@ public enum HealthKitConfig {
         hkQuantityTypeIdentifier: .dietaryFatTotal
       ),
       QuantityHealthKitMetric(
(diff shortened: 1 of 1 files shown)


chipmanaged:
> Please keep an eye out for the next testflight release, and confirm it works as expected.

@theospears I'm not in the TestFlight group. If you'd like to add me, I'll Slack you my email address.

chipmanaged:
This crashes for me when I try to connect dietary fibre and I see some obvious issues. I’ll take another run at it and see what I did wrong when I can test on device. Until then, this andd-on isn’t suitable for release.

theospears:
I was able to add dietary fiber and see it sync without issues with this PR applied

PR #709, opened 2026-01-07
Add workout type filtering for Apple Health workout minutes

theospears:
## Summary
- Allow users to filter which workout types are included when syncing workout minutes from Apple Health
- By default all workout types are included (including future iOS workout types the app doesn't know about)
- Users can select specific types organized into 7 categories (Cardio, Strength, Mind & Body, Racquet Sports, Team Sports, Individual Sports, Outdoor & Adventure)
- Configuration stored as `workout_types: [String]` using Swift case name identifiers for stability

## Test plan
- [ ] Verify workout configuration UI appears when selecting "Workout minutes" metric
- [ ] Verify "All Types" is the default selection
- [ ] Verify selecting specific workout types filters the preview data
- [ ] Verify configuration is saved and restored when editing an existing goal
- [ ] Verify "Daily Total" vs "Individual Workouts" toggle still works correctly
- [ ] Run unit tests: 21 new tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/BeeKit/HealthKit/WorkoutMinutesHealthKitMetric.swift
+++ b/BeeKit/HealthKit/WorkoutMinutesHealthKitMetric.swift
@@ -300,11 +300,23 @@ public class WorkoutMinutesHealthKitMetric: CategoryHealthKitMetric {
   // MARK: - Workout Filtering
 
   func filterWorkouts(_ workouts: [HKWorkout], config: [String: Any]) -> [HKWorkout] {
-    guard let typeIdentifiers = config["workout_types"] as? [String], !typeIdentifiers.isEmpty else {
+    guard let allowedTypes = allowedWorkoutTypes(from: config) else {
       return workouts  // No filter = all types
     }
-    let allowedTypes = Set(typeIdentifiers.compactMap { WorkoutActivityTypeInfo.find(byIdentifier: $0)?.activityType })
     return workouts.filter { allowedTypes.contains($0.workoutActivityType) }
+  }
+
+  private func allowedWorkoutTypes(from config: [String: Any]) -> Set<HKWorkoutActivityType>? {
+    guard let typeIdentifiers = config["workout_types"] as? [String], !typeIdentifiers.isEmpty else { return nil }
+    return Set(typeIdentifiers.compactMap { WorkoutActivityTypeInfo.find(byIdentifier: $0)?.activityType })
+  }
+
+  func workoutMatchesFilter(_ sample: HKSample, config: [String: Any]) -> Bool {
+    guard let workout = sample as? HKWorkout else { return false }
+    guard let allowedTypes = allowedWorkoutTypes(from: config) else {
+      return true  // No filter = all types
+    }
+    return allowedTypes.contains(workout.workoutActivityType)
   }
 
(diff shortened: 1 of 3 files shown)


BeeKit/HealthKit/WorkoutMinutesHealthKitMetric.swift
```diff
@@ -12,6 +29,210 @@ public class WorkoutMinutesHealthKitMetric: CategoryHealthKitMetric {
     super.init(humanText: humanText, databaseString: databaseString, category: category, hkSampleType: .workoutType())
   }
 
+  public override var hasAdditionalOptions: Bool { true }
+
+  // MARK: - Workout Type Definitions
+
+  public static let supportedWorkoutTypes: [WorkoutTypeInfo] = [
```
theospears: Let's move all this logic onto the WorkoutTypeInfo class instead

BeeKit/HealthKit/WorkoutMinutesHealthKitMetric.swift
```diff
@@ -27,25 +248,68 @@ public class WorkoutMinutesHealthKitMetric: CategoryHealthKitMetric {
...
         deadline: deadline,
         healthStore: healthStore,
         autodataConfig: autodataConfig
       )
     }
   }
-  func individualWorkoutDataPoints(days: Int, deadline: Int, healthStore: HKHealthStore) async throws -> [BeeDataPoint]
-  {
+
+  private func dailyAggregateDataPoints(
```
theospears: Can we reduce code duplication with the parent class, e.g. by having it take a predicate function parameter?

BeeSwiftTests/WorkoutConfigurationViewControllerTests.swift
```diff
@@ -0,0 +1,84 @@
...
+  func testConfigDailyAggregateTrue() {
+    let vc = createViewController()
+    // Default is daily aggregate (index 0)
+
+    let config = vc.getConfigParameters()
+
+    XCTAssertEqual(config["daily_aggregate"] as? Bool, true)
+  }
+
+  func testConfigDailyAggregateFalse() {
```
theospears: Update this test to actually adjust the control

BeeKit/HealthKit/WorkoutMinutesHealthKitMetric.swift
```diff
@@ -2,6 +2,289 @@ import Foundation
...
+      identifier: "flexibility",
+      displayName: "Flexibility",
+      category: .mindBody
+    ),
+    WorkoutActivityTypeInfo(
+      activityType: .mindAndBody,
+      identifier: "mindAndBody",
+      displayName: "Mind And Body",
+      category: .mindBody
+    ), WorkoutActivityTypeInfo(activityType: .dance, identifier: "dance", displayName: "Dance", category: .mindBody),
```
krugerk: Formatting? I suppose it has the same pattern as the ones to the start of mind and body but not to the ones directly around it.




Oh and this adds a new warning by calling deprecated `.dance`.

From iOS 14 it should be `.socialDance` and `.cardioDance` instead.

PR #713, opened 2026-01-11
Add column alignment to datapoint table

theospears:
## Summary
- Replaces single-label datapoint cell with three aligned columns: day, value, comment
- Calculates dynamic column widths based on actual content
- Day column uses max width and shows month prefix (e.g., "12/31") for datapoints from different months
- Value column uses 75th percentile width, with per-row overflow for outliers (long decimal values expand their row only)
- Right-aligns values for easier numeric comparison
- Fixes floating-point edge case in HH:MM minute calculation

## Test plan
- [ ] Navigate to a goal with recent datapoints - verify columns are aligned
- [ ] Test with mixed value lengths (short values like "0", long values like "123.456789")
- [ ] Test HH:MM format goals
- [ ] Test with datapoints from previous months - verify month prefix appears
- [ ] Test long comments - should truncate with ellipsis
- [ ] Verify ConfigureHKMetricViewController preview also shows aligned columns

🤖 Generated with [Claude Code](https://claude.com/claude-code)


BeeSwift/Components/DatapointTableViewController.swift
```diff
@@ -58,7 +65,12 @@ class DatapointTableViewController: UIViewController, UITableViewDelegate, UITab
     guard indexPath.row < datapoints.count else { return cell }
     let datapoint = datapoints[indexPath.row]
 
-    cell.datapointText = displayText(datapoint: datapoint)
+    let day = formatDay(datapoint: datapoint)
+    let value = formatValue(datapoint: datapoint)
+    let comment = datapoint.comment
+
+    let widths = columnWidths ?? DatapointColumnWidths(dayWidth: 20, valueWidth: 40)
+    cell.configure(day: day, value: value, comment: comment, columnWidths: widths)
```
theospears: Can we just pass the DataPoint into the cell rather than each attribute separately? And make the format methods public and static on the cell type.

BeeSwift/Components/DatapointTableViewController.swift
```diff
@@ -72,20 +84,67 @@ class DatapointTableViewController: UIViewController, UITableViewDelegate, UITab
...
+
+  private func calculatePercentileWidth(widths: [CGFloat], fallback: CGFloat) -> CGFloat {
+    guard !widths.isEmpty else { return fallback }
+
+    let sorted = widths.sorted()
+    let p75Index = min(sorted.count - 1, Int(ceil(Double(sorted.count) * 0.75)) - 1)
+    let p75Width = sorted[max(0, p75Index)]
+    let maxWidth = sorted.last!
+
+    if maxWidth <= 60 || maxWidth <= p75Width * 1.3 { return maxWidth } else { return p75Width }
```
theospears: Let's make the threshold 1.5x (and move to a constant)

BeeSwift/Components/DatapointTableViewController.swift
```diff
@@ -71,21 +81,4 @@ class DatapointTableViewController: UIViewController, UITableViewDelegate, UITab
 
     self.delegate?.datapointTableViewController(self, didSelectDatapoint: datapoint)
   }
-
-  func displayText(datapoint: BeeDataPoint) -> String {
```
krugerk: 💪
Locally I had branches with different approaches to this same issue though none of them were as many lines as this PR.
Looking forward to aligned recent data

theospears: Yeah, this is definitely a more “comprehensive” solution than I’d have written if I was writing all the code by hand.

krugerk: If the values contain a decimal point, how about having them line up on the decimal?

Here are two examples:
Bildschirmfoto 2026-01-11 um 23 29 54
Bildschirmfoto 2026-01-11 um 23 29 33

theospears: (Broadly I think avoiding causes of data points with too many decimals is the pragmatic fix)

PR #720, opened 2026-01-11
speeding up the gallery, a bit

krugerk:
## Summary
Upon returning to the gallery, especially after bring the app back into the foreground, scrolling the gallery can be very choppy.
This PR reduces some of the lagginess by moving json serializing off of the main thread and by prefetching the gallery's thumbnails.
RequestManager was made an actor and it was made to implement protocols to make testing/mocking easier.

## Validation
Used app in Simulator
logged out,
logged in,
browsed goal list (gallery),
opened goals (goal view),
edited and deleted datapoints from without the main app,
loaded settings

#588 mentioned a slow gallery.


BeeSwift/Gallery/GalleryViewController.swift
```diff
@@ -502,6 +503,18 @@ extension GalleryViewController: UICollectionViewDelegate {
...
 }
 
+// MARK: - Prefetching (using AlamofireImage's ImageDownloader)
+extension GalleryViewController: UICollectionViewDataSourcePrefetching {
+  func collectionView(_ collectionView: UICollectionView, prefetchItemsAt indexPaths: [IndexPath]) {
+    let urls = indexPaths.compactMap { indexPath -> URL? in
+      let goal = fetchedResultsController.object(at: indexPath)
+      return try? goal.thumbUrl.asURL()
+    }
+    let downloader = ImageDownloader.default
```
theospears: Currently `GoalImageView` has its own static `ImageDownloader`. Should we be either having this code use that downloader, or moving that to use the default one? I'm not sure if multiple instances share cache between them?

theospears (commented):
I agree this is needed! Added a few questions, but also happy to merge as-is if you think this is the best approach.

BeeKit/Managers/RequestManager.swift
```diff
@@ -62,9 +62,7 @@ public class RequestManager {
     ).validate().serializingData(emptyRequestMethods: [HTTPMethod.post]).response
 
     switch response.result {
-    case .success(let data):
-      let asJSON = try? JSONSerialization.jsonObject(with: data)
-      return asJSON
+    case .success(let data): return try await Task.detached { try JSONSerialization.jsonObject(with: data) }.value
```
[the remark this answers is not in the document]
krugerk: Switching it over to actor seems like a more complete solution.

BeeSwift/Gallery/GalleryViewController.swift
krugerk: Each will have its own cache. It makes sense that they both use the same one (some an item fetched and cached somewhere in the app is available elsewhere, in the cache).

BeeKit/Managers/RequestManaging.swift
```diff
@@ -0,0 +1,27 @@
...
+
+public protocol RequestManaging {
+  func get(url: String, parameters: [String: Any]?) async throws -> Any?
+  func put(url: String, parameters: [String: Any]?) async throws -> Any?
+  func post(url: String, parameters: [String: Any]?) async throws -> Any?
+  func delete(url: String, parameters: [String: Any]?) async throws -> Any?
+  func addDatapoint(urtext: String, slug: String, requestId: String?) async throws -> Any?
+}
+
+extension RequestManaging {
```
theospears: Can you explain what this is for? It seems like functions just calling themselves

krugerk: The protocol cannot have a default for the parameter and several places were calling them without specifying the parameter with a nil.

This here is a version that does (pass nil as a default) and allows then others to make calls without specifying also the parameters: nil for each and every call.

BeeSwift/Components/ImageDownloadService.swift
```diff
@@ -0,0 +1,10 @@
+// Part of BeeSwift. Copyright Beeminder
+
+import AlamofireImage
+
+final class ImageDownloadService {
```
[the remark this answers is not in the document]
krugerk: I understand the desire to have the singletons grouped like that. At the same time, currently this Service is in BeeSwift only and the ServiceLocator is in BeeKit. With the current split, the former is a UI only thing.
I will look into it.

BeeKit/Managers/RequestManager.swift
```diff
@@ -128,3 +98,33 @@ extension HTTPHeaders {
...
+  public func signedGET(url: String, parameters: [String: Any]?) async throws -> Any? {
+    let params = signedParameters(parameters)
+    return try await rawRequest(url: url, method: .get, parameters: params, headers: authenticationHeaders())
+  }
+  public func signedPOST(url: String, parameters: [String: Any]?) async throws -> Any? {
+    let params = signedParameters(parameters)
+    return try await rawRequest(url: url, method: .post, parameters: params, headers: authenticationHeaders())
+  }
+  fileprivate func signedParameters(_ params: [String: Any]?) -> [String: Any]? {
+    if params == nil { return params }
```
krugerk, line 112: guard

BeeKit/Managers/RequestManager.swift
```diff
@@ -128,3 +98,33 @@ extension HTTPHeaders {
...
+    var signed = params
+    var base = ""
+    var keys = Array(params!.keys)
```
krugerk, line 115: force

PR #744, opened 2026-02-24
an enum for the API Endpoints

krugerk:
## Summary
Trying a type for the API endpoint. Avoids much of the manual string manipulation and substring replacement.
Also provides a SSoT with regard to the url for an endpoint - the rest of the app does not need to deal with url paths, parameters, or http verbs.
Several places that had access to a username now pass it along rather than relying on the request manager to perform the substitution.


theospears:
I'm open to adding typing here, but I'd lean towards method-per-endpoint rather than enum. I don't love how the enum approach splits up the details of a given endpoint (parameters, method, signing etc) across multiple different helpers.

PR #760, opened 2026-04-05
migration path from b54 to post b74

### LINKED ISSUE #759: crash from App Store version to TestFlight version
The app crashes when moving from a logged in instance from the App Store version, 6.7 (54), to a recent TestFlight build, 6.8 (74).

The workaround is to remove the app (such that its data is also deleted) then reinstall. 

Although it might be expected/supported that users can install any TestFlight version over any App Store version, it is however expected that an update does not require removing and reinstalling the app.

https://github.com/beeminder/BeeSwift/compare/builds/iosbeta/54...builds/iosbeta/74

krugerk:
The current App Store version is [6.7 (54)](https://github.com/beeminder/BeeSwift/releases/tag/6.7). It contains just 'v1' of the Core Data BeeminderModel. Along the way, v2 and v3 were introduced in code and have not yet been released to the store. They have appeared however in dev (local) and TestFlight builds.

This PR restores v1 to how it was for the 6.7 (54) release and lets v2 (and then v3) take on the migration.

proposes to fix: #759


theospears:
I remember getting this lastUpdatedLocal/lastModifiedLocal stuff to migrate properly being a total nightmare.

BeeKit/Managers/GoalManager.swift
```diff
@@ -160,10 +160,10 @@ import SwiftyJSON
       let request = NSFetchRequest<Goal>(entityName: "Goal")
       request.predicate = NSPredicate(format: "id == %@", goalId)
       // TODO: Better error handling of failure here?
-      if let existingGoal = try! modelContext.fetch(request).first {
+      if let existingGoal = try? modelContext.fetch(request).first {
```
theospears: What's the error that is raised with the `try!` ?

krugerk: There are three `try`s in Swift: the normal one `do try catch`, `try?`, and `try!`.


--

let myVar = try? something.optional

does not throw but instead assigns myVar to nil or the non-optional thing but in both cases the type of myVar is Optional[ThatType]

--

let myVar = try! something.that.throws

the compiler will treat myVar as a non-optional because the try! tells the compiler 'yes we know this is stored as Optional[Something] and could throw but here we "guarantee" all will be well when accessed.

--




_Swift runtime failure: Unexpectedly found nil while unwrapping an Optional value_

krugerk: A similar place in code is 
https://github.com/beeminder/BeeSwift/blob/1e3a797b05a1b53c5cb65f0469398dc372ebaf9b/BeeKit/Managers/GoalManager.swift#L128

Where the refreshGoal method claims it `throws`, so callers can handle that, except for the method will crash the app with 

_Swift runtime failure: Unexpectedly found nil while unwrapping an Optional value_

when `existingObject(with: goalID)` cannot be for cast into a `Goal` rather than throwing.
Similarly the code there 
https://github.com/beeminder/BeeSwift/blob/1e3a797b05a1b53c5cb65f0469398dc372ebaf9b/BeeKit/Managers/GoalManager.swift#L130
says try this get which could result in an optional, stores responseObject as an optional, only to claim otherwise in 
https://github.com/beeminder/BeeSwift/blob/1e3a797b05a1b53c5cb65f0469398dc372ebaf9b/BeeKit/Managers/GoalManager.swift#L134
which again would crash the app when responseObject is actually nil.

krugerk: > What's the error that is raised with the `try!` ?

The code is at odds with itself here.

If there is an existing goal matching the id, grab it and update it.
If not, create a new one.

Is what we expect but the way the code is written with the try! it reads more like 

There is an existing goal matching the id, grab it and update it.
If not, we code is here to create a new one but we crashed already when the previous assumption (there is such a goal) failed.

theospears: If this was an accurate description of how the code worked, surely it would crash every time a new goal was created?

What actually happens is:
* `modelContext.fetch(request)` successfully returns 0 item list (`([BeeKit.Goal]) 0 values {}`) if there is no matching goal
* `.first` maps from a list to a `BeeKit.Goal?`
* The `if let` unwraps that optional

None of these involve an exception being caught by the `try!`. The `try!` instead catches some other error from `modelContext.fetch` which _is not_ no matching object found.

krugerk: Yeah, true. When it would throw, the try! becomes a runtime error.

And this method does not throw. So to make it more clear what went wrong when, then the exception could be directly caught and logged.

PR #794, opened 2026-06-12
Pin swift format 26.5

theospears:
This updates swift-format to 26.5 and also updates all files to match the new formatting performed by that formatter.

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -13,12 +13,12 @@ concurrency:
 
 jobs:
   pre-commit-check:
-    runs-on: macos-15
+    runs-on: macos-26
 
     steps:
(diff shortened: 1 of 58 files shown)


krugerk:
What is "swift-format 26.5"? https://github.com/swiftlang/swift-format refers to a latest version of 602.0.0 from Sept 2025. 
There is a .swift-format file in the repo but it was not touched in this PR. 
Instead, beside the formatting changes, I see in the pre-commit workflow was bumped to run on macOS 26 rather than macOS 15 and the Xcode version there as marked as 26.5.
(which brings it inline with 'runs-on: macOS-26' from e304ce0c4 and 'xcode-version: 26.5' from 0dea2c71f).
Well the wording on the PR description is a bit, odd.

PR #795, opened 2026-06-12
Gallery is only responsible for showing goals list

theospears:
## Summary
Previously the gallery handled a range of functions, including setup and the login screen. Move all of that elsewhere, and also only render the gallery when the user is logged in, so its only role is showing a list of goals.

## Validation
* [x] Launch the app in simulator and observe it loads

### DIFF
--- a/BeeSwift/MainCoordinator.swift
+++ b/BeeSwift/MainCoordinator.swift
@@ -49,6 +49,12 @@ class MainCoordinator {
     )
   }
   func start() {
+    navigationController.navigationBar.isTranslucent = false
+    navigationController.navigationBar.barStyle = .black
+    navigationController.navigationBar.tintColor = .white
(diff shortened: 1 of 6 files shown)


BeeSwift/MainCoordinator.swift
```diff
@@ -49,20 +51,23 @@ class MainCoordinator {
     )
   }
   func start() {
+    navigationController.navigationBar.isTranslucent = false
+    navigationController.navigationBar.barStyle = .black
+    navigationController.navigationBar.tintColor = .white
+    if currentUserManager.signedIn(context: viewContext) { showGallery() } else { showSignIn() }
```
krugerk, line 57: I find the formatting here unfortunate since the 'larger detail' is hidden further to the right (of the screen).

